PR #15, opened 2015-02-18
refs #8175 - Add linkRoutePattern support for dispatch router

stbenjam:
In order to actually connect an on-demand connector, we need to have support for `linkRoutePatterns`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ rvm:
   - 2.1.0
 script: bundle exec rake test
 env:
-  - PUPPET_VERSION="~> 2.7.0"
   - PUPPET_VERSION="~> 3.2.0"
   - PUPPET_VERSION="~> 3.3.0"
   - PUPPET_VERSION="~> 3.4.0"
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -3,13 +3,11 @@ language: ruby
 bundler_args: --without development
 before_install: rm Gemfile.lock || true
 rvm:
-  - 1.8.7
```
ehelms: We still need 1.8.7 support due to the installer using 1.8.7 no?

manifests/router/link_route_pattern.pp
```diff
@@ -0,0 +1,20 @@
...
+# Configure a link route pattern
+#
+# == Parameters
+#
+# $prefix::     Prefix to use
+#
+# $connector::  Connector for this link route pattern
+#
+define qpid::router::link_route_pattern(
+  $prefix    = 'queue.',
```
ehelms: Do prefixes always require a `.` to follow?

stbenjam: I think you could have a prefix like `queue.pulp` and have it match `queue.pulp_foo` and `queue.pulp_bar`, so I'd prefer to just leave it as a free form field.

.travis.yml
stbenjam: Right, sorry, looks like it was https://github.com/rspec/rspec-core/issues/1864 that was causing the failure, I pinned the rspec version to < 3.2.0...

stbenjam:
@ehelms Updated

ehelms:
ACK

PR #16, opened 2015-02-18
Address feedback from puppet forge

stbenjam:


ehelms:
@stbenjam what is your plan for putting this on puppet forge? This is a task across all our modules I have been considering for some time, once they were of the proper order. Do we have a way of knowing the "score" prior to publishing?

stbenjam:
@ehelms Not exactly, it's a combination of factors - puppet-lint, tests, complete metadata, etc. See http://puppetlabs.com/blog/new-ratings-puppet-forge-modules.   

My guess is qpid is in good shape, and we can address any feedback once 1.0 is published if there's any surprises but I think we should be 4.0 - 5.0. 


I don't really have any specific plan, I was thinking it could just be as we have time and the opportunity to do what we can.   I think generally the style of our modules is OK, the biggest thing is adopting semver, and making them more generic.  

So this one, puppet-servicewait and puppet-pulp are probably good candidates for the next ones - any others?

stbenjam:
@ehelms 4.2.  This PR should make it 5.0.

https://forge.puppetlabs.com/katello/qpid/scores

metadata.json
```diff
@@ -1,14 +1,23 @@
...
   "project_page": "https://github.com/Katello/puppet-qpid",
   "issues_url": "http://projects.theforeman.org/projects/katello/issues",
   "dependencies": [
-    { "name": "puppetlabs-stdlib", "version_requirement": ">= 3.0.0"},
-    { "name": "theforeman/concat_native", "version_requirement": ">= 1.3.0" }
+    { "name": "puppetlabs-stdlib", "version_requirement": ">= 3.0.0 < 5.0.0"},
+    { "name": "theforeman/concat_native", "version_requirement": ">= 1.3.0 < 2.0.0" }
+  ],
+  "requirements": [
+    { "name": "puppet", "version_requirement": ">= 2.7.0 < 4.0.0" }
```
ehelms: Since we are testing 3.2 and above - I'd think we'd want to reflect that here.

stbenjam:
@ehelms Updated

ehelms:
ACK

PR #17, opened 2015-02-20
refs #8175 - link router connector is optional

stbenjam:


manifests/router/link_route_pattern.pp
```diff
@@ -10,7 +10,7 @@
 #
 define qpid::router::link_route_pattern(
   $prefix    = 'queue.',
-  $connector = 'broker',
+  $connector = '',
```
ehelms: Better to use UNSET or undef instead of empty string? I am just basing this off of other places that do things like this.

stbenjam: `undef` is unreliable across puppet versions.  I could use UNSET, but the rest of the module uses empty strings.

stbenjam: Should I change them?

GregSutcliffe: Agreed undef isn't reliable. Once you accept you're testing a string, the contents of the default string become somewhat arbitrary. I'm not worried about it :)

ehelms:
ACK

PR #21, opened 2015-07-08
remove qpid-tools package

beav:
The qpid-tools package is a client-side package that is installed by other
modules and can cause conflicts if it is listed here. It is not needed for a
server install.


ehelms:
Makes sense, thanks @beav !

PR #23, opened 2015-07-16
Removes brackets around user_group parameter

dkliban:
The value defaults to [] already. No need to create an array of arrays.


ehelms:
ACK, thanks @dkliban !

PR #24, opened 2015-07-20
Revert "remove qpid-tools package"

stbenjam:
Reverts Katello/puppet-qpid#21

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -22,5 +22,5 @@ class qpid::params {
   $user = 'qpidd'
   $group = 'qpidd'
 
-  $server_packages = ['qpid-cpp-server', 'qpid-cpp-client', 'python-qpid-qmf',
-                      'python-qpid', 'policycoreutils-python'] }
+  $server_packages = ['qpid-cpp-server', 'qpid-cpp-client', 'python-qpid-qmf', 'python-qpid', 'policycoreutils-python', 'qpid-tools']
+}


stbenjam:
Looks like we need this package:

` /Stage[main]/Certs::Candlepin/Exec[create candlepin qpid exchange]: Could not evaluate: Could not find command 'qpid-config'`

```
[root@katello-centos6-bats vagrant]# yum whatprovides "*bin/qpid-config"
qpid-tools-0.30-4.el6.noarch : Management and diagnostic tools for Apache Qpid
```

ehelms:
Maybe we should put this on the puppet-certs then since that is what actually needs it and we might be less likely to break this in the future?

stbenjam:
If you're setting up a new qpid server, you want qpid-config to be able to create queues, etc.

If we want to not install qpid-tools by default, maybe we can refactor this to have some kind of flag like `install_tools`, but I really have a lot on my plate already with the Katello 2.3 to be the one to take this on.

I think it's proper to revert a commit that broke us.

stbenjam:
Another argument in favor of reverting: We're not great about semver in general in the modules, but this was a change that should've required a major version bump.  If you're using puppet-qpid 1.x you expect it to always install qpid-config, we can't just remove that behavior.

ehelms:
Sounds like the flag would be the proper way to go, if we revert this you'll need to do a release with the revert of the release. Our other option would be to lock the version in the Puppetfile and/or puppet certs for this module. I am good either way.

ehelms:
Package installation and requirement seems like a gray area, I doubt the puppet-certs module really even know that it was relying on this module to install that package. I can see your argument, but to me it's not really an interface that was being provided - if puppet-certs needs that package to exist then it should ensure it is available before running 'qpid-config'.

stbenjam:
I really disagree, puppet-qpid manages all things qpid, including providing the tools puppet-certs needed to work with it.  puppet-certs was mostly certainly relying on that behavior. :disappointed: 

Removing it requires more discussion, and a major version bump, I think this should be accepted and I'll release 1.0.4.

If @beav and others want the tools removed we can talk about introducing a flag that's false by default, but not until puppet-qpid 2.x.

ehelms:
ACK

stbenjam:
Thanks, released 1.0.3 https://forge.puppetlabs.com/katello/qpid

PR #25, opened 2015-07-21
missed parameter

larkit-ian:
Was playing with Geppetto and saw this.  The variable is called $router_mode inside router/params.pp


stbenjam:
Thanks @larkit-ian !

PR #26, opened 2015-08-04
fixes #11285 - require qpid-cpp-server-linearstore

stbenjam:


manifests/params.pp
```diff
@@ -22,5 +22,5 @@
   $user = 'qpidd'
   $group = 'qpidd'
 
-  $server_packages = ['qpid-cpp-server', 'qpid-cpp-client', 'python-qpid-qmf', 'python-qpid', 'policycoreutils-python', 'qpid-tools']
+  $server_packages = ['qpid-cpp-server', 'qpid-cpp-client', 'python-qpid-qmf', 'python-qpid', 'policycoreutils-python', 'qpid-tools', 'qpid-cpp-server-linearstore']
```
jlsherrill: I know older qpids (which katello doesn't use anymore) use qpid-cpp-server-store instead of linear-store.  Do we want to try to support that?   (The answer doesn't have to be yes)

stbenjam: Is it still in EPEL? I think it makes sense, this should probably be parameter and give the option of providing a store or not (same for things like installing the client, tools, etc).

jlsherrill: Looks like it is in epel 7:
https://dl.fedoraproject.org/pub/epel/7/x86_64/repoview/qpid-cpp-server-store.html

jlsherrill: but actually so is linear store.

ehelms: whats the status of this?

jlsherrill: I'm okay with merging as is (after rebase)

stbenjam:
Updated, the store being installed is configured, as is the package name, so we can support older qpids

ehelms:
ACK

stbenjam:
Thanks

PR #27, opened 2015-08-07
Support asymmetric routing in qpid dispatcher

stbenjam:
A new feature of dispatch router we're going to need to implement on the installer soon.

### DIFF
--- a/manifests/router/link_route_pattern.pp
+++ b/manifests/router/link_route_pattern.pp
@@ -6,10 +6,13 @@
 #
 # $prefix::     Prefix to use
 #
+# $direction:: Direction when using asymmetric routing
+#
 # $connector::  Connector for this link route pattern
 #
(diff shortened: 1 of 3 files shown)


manifests/router/link_route_pattern.pp
```diff
@@ -6,10 +6,13 @@
 #
 # $prefix::     Prefix to use
 #
+# $dir::        Direction when using asymmetric routing
```
ehelms: Could this be 'direction' since 'dir' means 'directory' in so many other languages and libraries?

stbenjam: I was on the fence but since you noticed it I'll use direction.

So far I had kept the parameter names from the config file verbatim, but this one is pretty poorly named.

stbenjam:
Updated

ehelms:
ACK

PR #28, opened 2015-09-04
Add qpid::tools package

ehelms:
Some module, such as Pulp, need to ensure the qpid_tools package
is present on a system. This creates a class to isolate qpid tools
to be used by this module and others.

### DIFF
--- a/manifests/tools.pp
+++ b/manifests/tools.pp
@@ -1,0 +1,8 @@
+# Class to ensure qpid tools package
+class qpid::tools {
+
+  package { 'qpid-tools':
+    ensure => 'installed',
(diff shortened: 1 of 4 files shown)


manifests/tools.pp
```diff
@@ -0,0 +1,8 @@
+# Class to ensure qpid tools package
+class qpid::tools {
+
+  package { 'qpid-tools':
+    ensure => 'installed'
```
stbenjam: Just need a `,`

ehelms:
Updated with comma

stbenjam:
ACK

PR #30, opened 2015-09-09
Change ssl_port to a string

ehelms:
The validate_re method expects a string and not an integer to be
pased to it. The integer value causes tests to fail on modules
that depend upon this qpid module.


stbenjam:
APT

PR #32, opened 2015-10-26
don't manage policycoreutils-python, as it is not related to qpid

cristifalcas:
And conflicts with other package that we manage separately. We are trying to keep this package updated on our own, even on machines that don't have this qpid package.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -24,7 +24,7 @@ class qpid::params {
   $user = 'qpidd'
   $group = 'qpidd'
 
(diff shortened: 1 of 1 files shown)


ehelms:
I wonder why this was included in the first place, could be a leftover from legacy or included due to Pulp. I'd like us to answer that question before we remove this.

ehelms:
Since I can't find the reason why this was here (and I went back as far as 2012 when we were writing all kinds of crazy modules with cross dependencies), I am OK with the change. I know Pulp uses it but it is required by their selinux package so I doubt we need to include this here. Thanks @cristifalcas !

PR #37, opened 2016-06-21
refs #15217 - puppet 4 support

stbenjam:


manifests/init.pp
```diff
@@ -71,7 +71,7 @@
 
   if $ssl {
     validate_bool($ssl_require_client_auth)
-    validate_re($ssl_port, '^\d+$')
+    validate_integer($ssl_port)
```
ekohl: ```ssl_port``` has ```type:string``` in the documentation which should be ```type:integer``` as well now.

stbenjam: Fixed

ekohl:
ACK

PR #41, opened 2016-10-14
Dispatch router log option is 'enable' not 'enabled'

stbenjam:
Otherwise:

Oct 14 07:57:49 qpid-asymmetric.example.com qdrouterd[25485]: Fri Oct 14 07:57:49 2016 ERROR (error) Python: Exception: Cannot load configuration file /etc/qpid-dispatch/qdrouterd.conf: org.apache.qpid.dispatch.log: Unknown attribute 'enabled' for 'org.apache.qpid.dispatch.log'


ehelms:
ACK

PR #50, opened 2017-02-06
Configure performance tuning settings for qpid

stbenjam:
When you have significant numbers of qpid or qdrouterd connections,
you need to set max-connections for qpid, and increase the number
of open files for qdrouterd.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,5 +2,6 @@ fixtures:
   repositories:
     stdlib: 'git://github.com/puppetlabs/puppetlabs-stdlib.git'
     concat: 'git://github.com/puppetlabs/puppetlabs-concat.git'
+    systemd: 'git://github.com/camptocamp/puppet-systemd.git'
   symlinks:
     qpid: "#{source_dir}"
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -32,6 +32,9 @@
 #
 # === Advanced parameters
 #
+# $max_connections::          Maximum number of connections to allow
+#                             type:integer
+#
 # $log_to_syslog::            Log to syslog or not
 #                             type:boolean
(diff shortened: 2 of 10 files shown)


stbenjam:
Note: this pulls in https://github.com/camptocamp/puppet-systemd to manage the systemd limits file.  It's not a big module and might come in handy later.  We could resort to managing the file manually if you'd prefer not to bring it in.

ehelms:
The only question I would have is whether you were wanting to backport this to any prior release that supported EL6? That would be the only limiting factor to using the systemd module (or just creating a separate patch for a backport).

stbenjam:
I don't think so, the module officially dropped el6 although it still probably works with it anyway, there's a `$::systemd` fact which says if it's a systemd distro and I do check that before setting the systemd-related setting.

stbenjam:
Although maybe we want to cherry pick the max-connections qpid setting to el6, I'll break this into 2 issues, to keep the systemd stuff separate.

stbenjam:
@ehelms I broke it into 2 commits, 32ef591 specifically addresses only the BZ to make it easier to cherry-pick.  e094084 has the systemd changes.

ehelms:
Do we need to consider finding a way to move `puppet-qpid` to the top of the installer or is the idea to leave this tuning for hiera config?

stbenjam:
Hiera for the moment, because of the cert generation and other interdependencies moving it and pulp to the top level is still a little challenging

stbenjam:
Ready for another look

ekohl:
I'm a bit hesitant in adding another dependency, but camptocamp does write good modules. The biggest risk I see is that they'll drop puppet 3 support earlier than we want to. Otherwise it looks good.

PR #51, opened 2017-03-07
Refs #18812 - Add event queue binding define

chris1984:
@stbenjam when you get a chance here is what I have so far, marking as wip.

### DIFF
--- a/manifests/bind_event.pp
+++ b/manifests/bind_event.pp
@@ -1,0 +1,35 @@
+# Define: qpid::bind_event
+#
+# This define binds the event queue with the correct messages
+#
+# === Parameters
+#
+# $queue::                      Name of the event queue
+#
+# $hostname::                   Set to localhost for qpid-config operations
+#
+# $port::                       Port that qpid is listening on
+#
+# $ssl_cert::                   SSL cert to use for qpid-config commands
+define qpid::bind_event(
+  $queue,
+  $hostname = 'localhost',
+  $port = 5671,
+  $ssl_cert = undef
+)
+{
+  if($ssl_cert) {
+    $ssl_option = "--ssl-certificate ${ssl_cert}"
+    $protocol = 'amqps'
+  } else {
+    $ssl_option = undef
+    $protocol = 'amqp'
+  }
+  exec { "bind queue to exchange and filter messages that deal with ${title}":
+    command   => "qpid-config ${ssl_option} -b ${protocol}://${hostname}:${port} bind event ${queue} ${title}",
+    onlyif    => "qpid-config ${ssl_option} -b ${protocol}://${hostname}:${port} exchanges event -r | grep ${title}",
+    path      => '/usr/bin',
+    require   => Service['qpidd'],
+    logoutput => true,
+  }
+}
--- a/spec/defines/bind_event.rb
+++ b/spec/defines/bind_event.rb
@@ -1,0 +1,36 @@
+require 'spec_helper'
+
+describe 'qpid::bind_event' do
+  let :title do
+    '*.*'
+  end
+
+  context 'without ssl_cert' do
+    let :params do
+      {
+        'queue' => 'myqueue',
+      }
+    end
+
+    it do
+      is_expected.to contain_exec('bind queue to exchange and filter messages that deal with *.*')
+        .with_command('qpid-config  -b amqp://localhost:5671 bind event myqueue *.*')
+        .with_onlyif('qpid-config  -b amqp://localhost:5671 exchanges event -r | grep *.*')
+    end
+  end
+
+  context 'with ssl_cert' do
+    let :params do
+      {
+        'queue'    => 'myqueue',
+        'ssl_cert' => '/path/to/cert.pem',
+      }
+    end
+
+    it do
+      is_expected.to contain_exec('bind queue to exchange and filter messages that deal with *.*')
+        .with_command('qpid-config --ssl-certificate /path/to/cert.pem -b ampqs://localhost:5671 bind event myqueue *.*')
+        .with_onlyif('qpid-config --ssl-certificate /path/to/cert.pem -b ampqs://localhost:5671 exchanges event -r | grep *.*')
+    end
+  end
+end


manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+# $hostname::                   FQDN of the host
+#
+# $ss_cert::                    SSL cert to use for qpid-config commands
+#
+# $candlepin_event_queue::      Name of the Candlepin event queue
+#
+# $title::                      Title of event
+define qpid::bind_event(
+include qpid::params
+  String $hostname = $::fqdn,
```
stbenjam: Looks like you're reading the Puppet 4 docs, at the moment don't use types like Strings on the variables, it's not supported in Puppet 3

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+  String $ssl_cert = '/etc/pki/katello/qpid_client_striped.crt',
```
stbenjam: This module isn't katello specific, there are non-katello users, leave this undef and pass it in, you should also try to support not having one, i.e. connecting over non-ssl

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+  String $candlepin_event_queue = 'katello_event_queue',
```
stbenjam: Ditto, don't call this "Candlepin" or specify a katello-specific name

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+  String $title
```
stbenjam: I don't think $title needs to be declared, it's built-in to a `define` already

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+#
+# $hostname::                   FQDN of the host
+#
+# $ss_cert::                    SSL cert to use for qpid-config commands
+#
+# $candlepin_event_queue::      Name of the Candlepin event queue
+#
+# $title::                      Title of event
+define qpid::bind_event(
+include qpid::params
```
stbenjam: Don't need this

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+  String $hostname = $::fqdn,
+  String $ssl_cert = '/etc/pki/katello/qpid_client_striped.crt',
+  String $candlepin_event_queue = 'katello_event_queue',
+  String $title
+) {
+    exec { "bind katello entitlements queue to qpid exchange messages that deal with ${title}":
+    command   => "qpid-config --ssl-certificate ${ssl_cert} -b 'amqps://localhost:5671' bind event ${candlepin_event_queue} ${title}",
```
stbenjam: I'd try to make localhost and 5671 parameterized

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+    onlyif    => "qpid-config --ssl-certificate ${ssl_cert} -b 'amqps://localhost:5671' exchanges event -r",
```
stbenjam: This should probably have a grep for checking the binding

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+    path      => '/usr/bin',
+    require   => Service['qpidd'],
+    logoutput => true,
+  }
+}
+
+qpid::bind_event {'entitlement.created'}
```
stbenjam: These would go in puppet-katello

stbenjam:
@chris1984  The general idea is right, looks like a good start

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
+# Define: qpid::bind_event
+#
+# This define binds the event queue in katello with the correct messages from candlepin to filter out uneeded messages
+#
+# === Parameters
+#
+# $hostname::                   FQDN of the host
+#
+# $ss_cert::                    SSL cert to use for qpid-config commands
```
ekohl: Typo here: ```$ssl_cert```

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
+# Define: qpid::bind_event
+#
+# This define binds the event queue in katello with the correct messages from candlepin to filter out uneeded messages
```
ehelms: Drop references to Katello and Candlepin throughout. Try to think of `puppet-qpid` as a module that knows how to configure Qpid and that other modules can configure it to their needs. So you want to strive for making and referencing generic parameters and concepts specific to qpid alone.

manifests/bind_event.pp
```diff
@@ -0,0 +1,27 @@
...
+# $port::                       Port that qpid is listening on
+define qpid::bind_event(
+  $hostname = 'localhost',
+  $ssl_cert = undef,
+  $queue = 'event_queue',
+  $port = '5671'
+) {
+    exec { "bind entitlements queue to qpid exchange messages that deal with ${title}":
+    command   => "qpid-config --ssl-certificate ${ssl_cert} -b 'amqps://${hostname}:${port}' bind event ${queue} ${title}",
+    onlyif    => "qpid-config --ssl-certificate ${ssl_cert} -b 'amqps://${hostname}:${port}' exchanges event -r | grep ${title}",
```
stbenjam: One improvement we could make here is supporting non-ssl connections.  It currently assumes that it's always SSL. You can do something like this above the exec

```puppet
if($ssl_cert) {
  $ssl_option = "--ssl-certificate ${ssl_cert}"
  $protocol = "ampqs"
} else {
  $ssl_option = ""
  $protocol = "amqp"
}
```

and then put `${protocol}` and `${ssl_option}` in the right places in command

chris1984:
@ehelms @stbenjam can I get another review please, I made all the suggested corrections with removing any references to Katello/Candlepin and fixed up the syntax so the tests would pass.

manifests/bind_event.pp
stbenjam: Also lint didn't flag this, but this should all be indented under the exec

```puppet
exec {'foo':
  bar => 'baz',
}
```

manifests/bind_event.pp
```diff
@@ -0,0 +1,27 @@
...
+# $ssl_cert::                    SSL cert to use for qpid-config commands
+#
+# $queue::                      Name of the event queue
+#
+# $port::                       Port that qpid is listening on
+define qpid::bind_event(
+  $hostname = 'localhost',
+  $ssl_cert = undef,
+  $queue = 'event_queue',
+  $port = '5671'
```
stbenjam: Just make this an integer, quoting it makes it a string

stbenjam: (Also, If you take the non-SSL suggestion, the default should be 5672)

manifests/bind_event.pp
```diff
@@ -0,0 +1,27 @@
+# Define: qpid::bind_event
+#
+# This define binds the event queue with the correct messages
+#
+# === Parameters
+#
+# $hostname::                   Set to localhost for qpid-config operations
+#
+# $ssl_cert::                    SSL cert to use for qpid-config commands
```
stbenjam: This description isn't aligned with the others

chris1984:
@stbenjam I was getting an issue with puppet lint when run test locally saying: variable assigned to the empty string on line xx with single quotes. With "" it was not happy so tried '' and then got the above so I moved it to undef, not sure if that is the correct way. Fixed the other changes though.

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+#
+# $port::                       Port that qpid is listening on
+define qpid::bind_event(
+  $hostname = 'localhost',
+  $ssl_cert = undef,
+  $queue = 'event_queue',
+  $port = 5671
+) {
+  if($ssl_cert) {
+  $ssl_option = "--ssl-certificate ${ssl_cert}"
```
ekohl: Indenting is off here

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+#
+# $ssl_cert::                   SSL cert to use for qpid-config commands
+#
+# $queue::                      Name of the event queue
+#
+# $port::                       Port that qpid is listening on
+define qpid::bind_event(
+  $hostname = 'localhost',
+  $ssl_cert = undef,
+  $queue = 'event_queue',
```
ekohl: It's also fine not to have a default. In this case I think that makes sense since there is nothing that works for most people.

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+  $port = 5671
+) {
+  if($ssl_cert) {
+  $ssl_option = "--ssl-certificate ${ssl_cert}"
+  $protocol = 'ampqs'
+} else {
+  $ssl_option = undef
+  $protocol = 'amqp'
+}
+    exec { "bind entitlements queue to qpid exchange messages that deal with ${title}":
```
ekohl: I'd leave out the entitlements here.

stbenjam:
@chris1984 Ah ok, undef is fine then.  Looks good to me, we should be good to go once @ekohl's feedback is addressed.

chris1984:
@stbenjam @ekohl thanks again for the help with this, let me know how the latest revision looks. This was definitely a fun journey into Puppet land.

ekohl:
The following test can be used to verify it does what it's supposed to do. Note that I used the ampqs protocol in the test so it passes with the current code.

```ruby
require 'spec_helper'

describe 'qpid::bind_event' do
  let :title do
    '*.*'
  end

  context 'without ssl_cert' do
    let :params do
      {
        'queue' => 'myqueue',
      }
    end

    it do
      is_expected.to contain_exec('bind queue to exchange and filter messages that deal with *.*')
        .with_command('qpid-config  -b amqp://localhost:5671 bind event myqueue *.*')
        .with_onlyif('qpid-config  -b amqp://localhost:5671 exchanges event -r | grep *.*')
    end
  end

  context 'with ssl_cert' do
    let :params do
      {
        'queue'    => 'myqueue',
        'ssl_cert' => '/path/to/cert.pem',
      }
    end

    it do
      is_expected.to contain_exec('bind queue to exchange and filter messages that deal with *.*')
        .with_command('qpid-config --ssl-certificate /path/to/cert.pem -b ampqs://localhost:5671 bind event myqueue *.*')
        .with_onlyif('qpid-config --ssl-certificate /path/to/cert.pem -b ampqs://localhost:5671 exchanges event -r | grep *.*')
    end
  end
end
```

chris1984:
@ekohl is this all set to go, or do you want me to add that test?

chris1984:
Local tests passing with the new one:

/home/chrobert/.rvm/rubies/ruby-2.1.5/bin/ruby -I/home/chrobert/.rvm/gems/ruby-2.1.5/gems/rspec-core-3.5.4/lib:/home/chrobert/.rvm/gems/ruby-2.1.5/gems/rspec-support-3.5.0/lib /home/chrobert/.rvm/gems/ruby-2.1.5/gems/rspec-core-3.5.4/exe/rspec --pattern spec/\{aliases,classes,defines,unit,functions,hosts,integration,type_aliases,types\}/\*\*/\*_spec.rb --color
..................................................................

Deprecation Warnings:

Using `should` from rspec-expectations' old `:should` syntax without explicitly enabling the syntax is deprecated. Use the new `:expect` syntax or explicitly enable `:should` with `config.expect_with(:rspec) { |c| c.syntax = :should }` instead. Called from /home/chrobert/Dropbox/git/puppet-qpid/spec/classes/qpid_client_config_spec.rb:17:in `block (5 levels) in <top (required)>'.


If you need more of the backtrace for any of these deprecations to
identify where to make the necessary changes, you can configure
`config.raise_errors_for_deprecations!`, and it will turn the
deprecation warnings into errors, giving you the full backtrace.

1 deprecation warning total

Finished in 5.05 seconds (files took 1.67 seconds to load)
66 examples, 0 failures

chris1984:
@ekohl tests are failing after removing that, can I put it back?

stbenjam:
Just change the order

```
manifests/bind_event.pp:16:WARNING: optional parameter listed before required parameter
```

chris1984:
@stbenjam @ekohl can I get another look when you get time please, all corrections made and tests are passing.

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+#
+# === Parameters
+#
+# $ssl_cert::                   SSL cert to use for qpid-config commands
+#
+# $queue::                      Name of the event queue
+#
+# $hostname::                   Set to localhost for qpid-config operations
+#
+# $port::			Port that qpid is listening on
```
stbenjam: I'd suggest reviewing diffs yourself before asking for reviews so we have less back-and-forth. Indentation is wrong here.

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+define qpid::bind_event(
+  $ssl_cert,
+  $queue,
+  $hostname = 'localhost',
+  $port = 5671
+) {
+  if($ssl_cert) {
```
stbenjam: Indentation is kind of weird here, the body of the define should have all the members at the same level of indentation, e.g. this if block should be at the same leve as exec

manifests/bind_event.pp
```diff
@@ -0,0 +1,34 @@
...
+#
+# $ssl_cert::                   SSL cert to use for qpid-config commands
+#
+# $queue::                      Name of the event queue
+#
+# $hostname::                   Set to localhost for qpid-config operations
+#
+# $port::			Port that qpid is listening on
+define qpid::bind_event(
+  $ssl_cert,
```
ekohl: ```ssl_cert``` needs to be undef so it's optional.

chris1984:
puppet-lint is not complaining about indentation with the if statement moved to the exec block. This is telling me to move it back https://github.com/relud/puppet-lint-strict_indent-check?

stbenjam:
Also, I opened a PR to add the strict indent check to our checks, so we officially check for 2 space indents. https://github.com/Katello/foreman-installer-modulesync/pull/19

stbenjam:
Thanks everyone

PR #52, opened 2017-03-22
Re-factor bind_event to the generic bind action

ehelms:
Moves away from having a bind action to a specific exchange and allows
using bind as a generic type as defined by qpid. This change now requires
exchange to be specified.

Ref: https://qpid.apache.org/releases/qpid-0.32/cpp-broker/book/chapter-Managing-CPP-Broker.html

### DIFF
--- a/manifests/config/bind.pp
+++ b/manifests/config/bind.pp
@@ -1,21 +1,21 @@
-# Define: qpid::bind_event
+# Define: qpid::config::bind
 #
-# This define binds the event queue with the correct messages
+# This define binds the queue with the correct messages
 #
 # === Parameters
 #
-# $queue::                      Name of the event queue
+# $queue::                      Name of the queue
 #
 # $exchange::                   Name of the exchange the queue is on
 #
-# $hostname::                   Set to localhost for qpid-config operations
+# $hostname::                   Hostname of the qpid broker
 #
 # $port::                       Port that qpid is listening on
 #
 # $ssl_cert::                   SSL cert to use for qpid-config commands
-define qpid::bind_event(
+define qpid::config::bind(
   $queue,
-  $exchange = 'event',
+  $exchange,
   $hostname = 'localhost',
   $port = undef,
   $ssl_cert = undef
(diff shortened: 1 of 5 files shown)


ehelms:
This shouldn't be merged until other areas of code using this (and there are some re-factoring on going) are updated.

ekohl:
I was thinking of introducing a wrapper around ```qpid-config```. Maybe a define named ```qpid::config_command``` that would take similar arguments. The bind could build on top of that. Given we only have one use of bind, maybe even that isn't needed. We could have both in the code base while we move from bind_event to the new structure.

Klaas-:
do we really want a require for Service['qpidd']? shouldn't this be conditional if the qpidd is not on the same server as for example katello or candlepin where the exchanges are created? Or do we want to handle all qpid-config commands on the server where qpidd is runinng?

ekohl:
@ehelms https://github.com/Katello/puppet-qpid/pull/53 is the solution I proposed.

@Klaas- that's a good point. We could add a collector to make it optional. Passing in the URL suggests you can run this remote if you want to.

ehelms:
Updated with the slight modification to keep this generic

ehelms:
Updated:

 * Fixes testing for new bind type
 * Adds exchange type and testing
 * Moves those types under config/ and ::config::

manifests/config/bind.pp
```diff
@@ -1,10 +1,10 @@
-# Define: qpid::bind_event
+# Define: qpid::config::bind
 #
 # This define binds the event queue with the correct messages
```
ekohl: Is this still correct?

manifests/config/exchange.pp
```diff
@@ -0,0 +1,29 @@
+# Define: qpid::config::exchange
+#
+# This define ensures an exchange exists
+#
+# === Parameters
+#
+# $exchange::                   Name of the exchange the queue is on
```
ekohl: The queue part can be left out?

manifests/config/exchange.pp
```diff
@@ -0,0 +1,29 @@
...
+#
+# $hostname::                   Set to localhost for qpid-config operations
```
ekohl: Not sure this is correct.

PR #53, opened 2017-03-24
Introduce qpid::config_cmd as qpid-config wrapper

ekohl:


manifests/bind_event.pp
```diff
@@ -6,30 +6,26 @@
...
 #
 # $port::                       Port that qpid is listening on
 #
 # $ssl_cert::                   SSL cert to use for qpid-config commands
 define qpid::bind_event(
   $queue,
+  $exchange = 'event',
   $hostname = 'localhost',
-  $port = 5671,
+  $port = undef,
```
ehelms: Why not default this to the standard qpid port?

ehelms:
I like this proposed addition and would prefer to merge it then re-factor my change on top of this structure.

manifests/bind_event.pp
ekohl: Because the port is determined in the wrapper. That means it automatically selects the correct port for SSL and non-SSL.

ehelms: Touche!

PR #55, opened 2017-03-28
config_cmd: optional qpidd, documentation, tests

ekohl:
* Makes the dependency on Service[qpidd] optional
* Documents the $unless parameter
* Removes the hardcoded defaults from candlepin::config::*
* Adds tests

### DIFF
--- a/manifests/config_cmd.pp
+++ b/manifests/config_cmd.pp
@@ -9,11 +9,18 @@
 # $onlyif::                     The qpid-config command including parameters to
 #                               check if $command should be executed
 #
+# $unless::                     The qpid-config command including parameters to
+#                               check if $command should not be executed.
+#                               Ignored if $onlyif is specified
+#
 # $hostname::                   The hostname to connect to
 #
 # $port::                       Port that qpid is listening on
 #
 # $ssl_cert::                   SSL cert to use for qpid-config commands
+#
+# $ssl_key::                    SSL key to use for qpid-config commands
+#
 define qpid::config_cmd (
   $command,
   $onlyif = false,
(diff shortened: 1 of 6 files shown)


manifests/config_cmd.pp
```diff
@@ -30,21 +34,22 @@
     $base_cmd = "qpid-config -b amqp://${hostname}:${_port}"
   }
 
+  Service <| tag == 'qpidd' |> ->
```
ehelms: If I understand correctly, this will make it usable where the qpid-client and qpid server are on separate hosts?

ekohl: That is correct. It does look like the exact syntax I used (chain to an in with resources inside it) is puppet 4 only so I'll modify it to work on puppet 3 as well. You can see it in the tests as well where I test both with an without qpid.

ehelms:
I think qpid_config needs to accept a key and have that passed in along with cert:

````
# [ WARN 2017-03-28 18:47:44 verbose]  /Stage[main]/Katello::Qpid/Qpid::Config_cmd[create katello entitlements queue]/Exec[qpid-config create katello entitlements queue]/returns: Failed: ConnectError: [SSL] PEM lib (_ssl.c:2521)
# [ERROR 2017-03-28 18:47:44 verbose]  qpid-config --ssl-certificate /etc/pki/katello/certs/centos7-bats-ci.example.com-qpid-broker.crt -b amqps://localhost:5671 add queue katello_event_queue --durable returned 1 instead of one of [0]
```

Do you want to change that here or I can do a separate PR.

ekohl:
Updated

ehelms:
Bind and exchange defines should also get key options?

ekohl:
Good point, updated again :)

ehelms:
Looks good once tests are passing

ekohl:
Now with passing tests

PR #56, opened 2017-03-29
Contain classes

ekohl:


ekohl:
This introduces a dependency cycle:

```
Cert[proxy-nightly.example.com-qpid-client-cert] => Class[Certs::Qpid_client] => Class[Qpid] => Class[Qpid::Config] => Group[qpidd] => File[/etc/pki/katello/private/proxy-nightly.example.com-qpid-broker.key] => Certs::Keypair[qpid] => Class[Certs::Qpid] => Class[Certs::Qpid_client] => Cert[proxy-nightly.example.com-qpid-client-cert]
```

ehelms:
This indeed fixed the issue that occured during a proxy installation for me. Once tests are passing -- ACK

manifests/init.pp
```diff
@@ -93,8 +93,9 @@
     validate_absolute_path($ssl_cert_db, $ssl_cert_password_file)
   }
 
-  class { '::qpid::install': } ~>
-  class { '::qpid::config': } ~>
-  class { '::qpid::service': } ->
-  Class['qpid']
+  include ::qpid::install
```
timogoebel: Why not contain all of the classes?

If I want to do a `require => Class['qpid']` to ensure the package is installed, this still suffers from the same issue.

ekohl: That results into a dependency cycle. And since we contain the service which in turn requires the package that should all work out.

ekohl: It does mean that if you want something to happen before the install, you're not sure it is contained.

ekohl:
Added 2 commits to make puppet-lint happy and while I was at it the specs.

manifests/config.pp
```diff
@@ -7,7 +7,8 @@
 
   group { $qpid::group:
     ensure => present,
-  } ->
```
ekohl: This dependency should be safe to drop since https://docs.puppet.com/puppet/latest/type.html#user-description states:
> Autorequires: If Puppet is managing the user’s primary group (as provided in the gid attribute) or any group listed in the groups attribute then the user resource will autorequire that group. If Puppet is managing any role accounts corresponding to the user’s roles, the user resource will autorequire those role accounts.

manifests/client.pp
```diff
@@ -14,7 +14,9 @@
   $client_packages         = $qpid::client::params::client_packages,
 ) inherits qpid::client::params {
 
-  class { '::qpid::client::install': } ~>
-  class { '::qpid::client::config': }
+  include ::qpid::client::install
+  include ::qpid::client::config
+
+  Class['qpid::client::install'] ~> Class['qpid::client::config']
```
ekohl: I did not like the ~> on the newline so I changed it to this format. Maybe we should standardize on something.

PR #60, opened 2017-05-10
Allow puppetlabs-concat 4.0.0

ekohl:


chris1984:
@ekohl I see no one reviewed this one yet, any negative effects of bringing this new version of concat in?

ekohl:
Not really. In the foreman modules we've seen a bug in librarian-puppet where we need to pin it at foreman-installer level because it can't figure out that other modules only allow 2.x but we can work around that.

PR #62, opened 2017-05-11
Fixes #19514 - Add session-max-unacked option

chris1984:


manifests/init.pp
```diff
@@ -22,6 +22,8 @@
 #
 # $ssl_require_client_auth::  Require client SSL authentication
 #
+# $session_unacked:           buffer if the broker has a large number of sessions and the memory overhead is a problem
```
stbenjam: This is missing from the class parameters below

sean797:
https://bugzilla.redhat.com/show_bug.cgi?id=1367735 Also mentions that setting 'wcache-page-size=' can help. Should we allow for that as well ? 
(Doesn't have to be done in this PR)

chris1984:
@sean797 +1 to adding it, I'll make another PR once @stbenjam helps me get this one going.

manifests/init.pp
```diff
@@ -53,6 +55,7 @@
   String $server_store_package = $::qpid::params::server_store_package,
   Boolean $ssl = $::qpid::params::ssl,
   Integer[0, 65535] $ssl_port = $::qpid::params::ssl_port,
+  Integer $session_unacked = $::qpid::params::session_unacked,
```
ekohl: You could use ```Integer[0]``` to set a lower bound of 0, assuming negative integers are not supported.

chris1984:
@ekohl can I get another review please? I fixed the init.pp with Integer [0] also I got the spec tests green.

stbenjam:
If I'm reading the [upstream issue correctly,](https://issues.apache.org/jira/browse/QPID-7677) this is only in qpid 1.37 or newer.  The latest in EPEL is 1.36.

How does older qpid respond to unknown options? I think maybe it should be made optional, and make the default value undef.

chris1984:
@stbenjam It looks on 6.2 it causes an error with bad argument:

Starting Qpid AMQP daemon: 2017-05-15 06:22:15 [Broker] critical Unexpected error: Error in configuration file /etc/qpid/qpidd.conf: Bad argument: |session-max-unacked=10|

Use --help to see valid options

I will change to optional and value = undef and have the config be nil unless a value is presented.

pmoravec:
Unknown options means `qpidd` wont start, as @chris1984 described.

One option is to check `qpid-cpp-server` version and use that option only when it is >=1.37. BUT that won't work ideally in case the [QPID-7677](https://issues.apache.org/jira/browse/QPID-7677) is backported in downstream to a 1.34 or 1.36 version.

Could there be a test (I am not such an expert on puppet) that if `qpidd --help | grep session-max-unacked` is nonempty, then (and only then) add this value to `/etc/qpid/qpidd.conf`?

ekohl:
> Could there be a test (I am not such an expert on puppet) that if qpidd --help | grep session-max-unacked is nonempty, then (and only then) add this value to /etc/qpid/qpidd.conf?

You could add a fact for this, but that would need to be present before puppet starts to use it. Since it's possible that ```qpid-cpp-server``` is installed during the puppet run, it's very tricky. Given we haven't really guaranteed idempotency in katello-installer (though that's high on my wishlist) running twice would also have serious downsides.

chris1984:
@stbenjam updated the PR when you get a chance to look at it.

stbenjam:
Looks ok to me.  Version facts are problematic, this change is fine, we just need a puppet-foreman_proxy_content change to set the max unacked once we're on 1.37.

I'd also prefer to have some test coverage, can you add a test like this one?

https://github.com/Katello/puppet-qpid/blob/master/spec/classes/qpid_config_spec.rb#L24-L39

chris1984:
@stbenjam added the test similar to what you posted, from the quick recap on irc:

still having some issues with the test when i make it, ill push up the changes for you to review. it seems to not actually be setting the var so the tests fail since the element does not contain the value. i could be doing something wrong too but pretty much looks like all the other tests

templates/qpidd.conf.erb
```diff
@@ -44,3 +44,6 @@ interface=<%= scope['qpid::interface'] %>
 <% unless [nil, :undefined, :undef, ''].include?(scope['qpid::max_connections']) -%>
 max-connections=<%= scope['qpid::max_connections'] %>
 <% end %>
+<% unless [nil, :undefined, :undef, ''].include?(scope['qpid::params::session_unacked']) -%>
```
stbenjam: This should be `qpid::session_acked`, not the value in qpid::params.  Params is a pattern to hold default values, it'll always be undef.

chris1984:
@stbenjam all tests are passing :) thanks again for the help.

stbenjam:
Thanks @chris1984 !

PR #65, opened 2017-06-05
Refs #19929 - Add wcache-page-size to installer

chris1984:


manifests/init.pp
```diff
@@ -28,6 +28,8 @@
 #
 # $max_connections::          Maximum number of connections to allow
 #
+# $wcache_size::              The size (in KB) of the pages in the write page cache
```
ehelms: Since the config parameter is named `wcache_page_size` can we call this parameter the same for 1:1 mapping? Leaves no room for guess work :)

chris1984:
@ehelms updated with wcache_page_size for var name, if you can review again please.

PR #71, opened 2017-08-11
Improve qpid::config::* defines

ekohl:
See individual commits for exact changes.

### DIFF
--- a/manifests/config/bind.pp
+++ b/manifests/config/bind.pp
@@ -25,7 +25,8 @@ define qpid::config::bind(
   $ssl_key = undef,
 )
 {
-  qpid::config_cmd { "bind queue to exchange and filter messages that deal with ${title}":
+  $cmd = "bind queue to exchange and filter messages that deal with ${title}"
+  qpid::config_cmd { $cmd:
     command  => "bind ${exchange} ${queue} ${title}",
     unless   => "exchanges ${exchange} -r | grep ${title}",
     hostname => $hostname,
@@ -33,4 +34,7 @@ define qpid::config::bind(
     ssl_cert => $ssl_cert,
     ssl_key  => $ssl_key,
   }
+
+  Qpid::Config::Exchange <| exchange == $exchange |> -> Qpid::Config_cmd[$cmd]
+  Qpid::Config::Queue <| queue == $queue |> -> Qpid::Config_cmd[$cmd]
 }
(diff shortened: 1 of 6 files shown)


manifests/config/bind.pp
```diff
@@ -25,12 +25,16 @@
...
     command  => "bind ${exchange} ${queue} ${title}",
     unless   => "exchanges ${exchange} -r | grep ${title}",
     hostname => $hostname,
     port     => $port,
     ssl_cert => $ssl_cert,
     ssl_key  => $ssl_key,
   }
+
+  Qpid::Config::Exchange <| exchange == $exchange |> -> Qpid::Config_cmd[$cmd]
+  Qpid::Config::Queue <| queue == $queue |> -> Qpid::Config_cmd[$cmd]
```
ehelms: So this adds a requirement that the queue and exchange were created/managed by Puppet somewhere within the puppet apply that includes this module, if I read correctly?

manifests/config/queue.pp
```diff
@@ -0,0 +1,33 @@
+# Define: qpid::config::queue
+#
+# This define ensures an queue exists
```
ehelms: s/a/an

manifests/config/queue.pp
```diff
@@ -0,0 +1,33 @@
...
+define qpid::config::queue(
+  $queue = $title,
+  $hostname = undef,
+  $port = undef,
+  $ssl_cert = undef,
+  $ssl_key = undef,
+)
+{
+  qpid::config_cmd { "ensure queue ${queue}":
+    command  => "add queue ${queue} --durable",
```
ehelms: Should durable be an option here?

manifests/config/bind.pp
ekohl: Correct. It also works if they aren't created/managed (at least, there's tests for that and I don't see any errors)

manifests/config/queue.pp
ekohl: It's what puppet-katello does. I also thought about parametrizing this and wondered what the purpose of adding non-durable queues using puppet is.

manifests/config/queue.pp
ekohl: Result of ```s/exchange/queue```. Fixing.

ekohl:
Rebased and s/an/a

PR #83, opened 2018-01-16
fixes #22289 - update for dispatch router 1.0

stbenjam:
Updated for qpid 1.0, still needs testing.  Also looking through the docs to see if we need to change anything else.


manifests/router.pp
```diff
@@ -17,13 +15,15 @@
 #
 # $router_packages:: The package to be installed
 #
+# $worker_threads::  Number of worker threads
```
stbenjam: Not sure why this wasn't exposed as a class param

manifests/router/listener.pp
```diff
@@ -18,7 +18,7 @@
 #
 define qpid::router::listener(
   $ssl_profile    = undef,
-  $addr           = '0.0.0.0',
+  $host           = '0.0.0.0',
```
ekohl: I'm thinking of not renaming these initially because that requires changes in https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/dispatch_router.pp while dropping the container doesn't hurt since it's not overridden.

stbenjam: Yea I know, but there's other renames, for example linkRoutePattern is now linkRoute and takes different params too.

If @ehelms could pin the 3.5 RPM's to older dispatch router then we could take care of all the upcoming deprecations in one shot and make puppet-qpid 4.x ready for dispatch router 1.0 and do that in 3.6?

I can also make a smaller scope change and just fix the container for 3.5, but who knows when they'll break us again by removing the other deprecated items

ekohl: We don't have to rename the parameters if they continue to accept the same values. It'll mean we need to do a 4.0 and a 5.0 release but that can be worth it.

stbenjam: I guess, it's nicer when things match the qpid docs.  I'll open a smaller PR with the minimal changes for 1.0.

stbenjam:
Goes with https://github.com/theforeman/puppet-foreman_proxy_content/pull/155

@ekohl Since we need to change the katello use of this module anyway, think we can change the class and variable names so they match qpid config? (e.g. addr -> host) or would you rather leave the module more compatible with how it used to do things?

ehelms:
Given this is a major version bump, I am all for ensuring consistency with the actual project being managed. Backporting this for Katello 3.5 and 3.6 is going to require modifying `puppet-foreman_proxy_content` in either case right? 

Do both cases require installer updates as well?

stbenjam:
Yea we'll need changes to puppet-foreman_proxy_content in either case, so seems to make sense to do all the breaking in one go.  Installer shouldn't need any changes, puppet-qpid is wrapped by foreman_proxy_content.

ekohl:
My previous position was based on not having to change the proxy side of it but since that's no longer true, let's break it all the way.

stbenjam:
Ready to look again, tested on 3.5 with theforeman/puppet-foreman_proxy_content#155.

manifests/router/connector.pp
```diff
@@ -17,12 +17,12 @@
...
-  $addr         = '0.0.0.0',
-  $port         = 5672,
-  $sasl_mech    = 'ANONYMOUS',
-  $role         = undef,
-  $ssl_profile  = undef,
-  $idle_timeout = undef,
+  String $host = '127.0.0.1',
+  Integer[0, 65535] $port = 5672,
+  Optional[String] $sasl_mech = 'ANONYMOUS',
+  Optional[String] $role = undef,
```
ekohl: From the docs:
> role (One of [normal, inter-router, route-container], default=normal)

So can be ```Optional[Enum['normal', 'inter-router', 'route-container']]```

manifests/router/listener.pp
```diff
@@ -17,12 +17,12 @@
...
-  $addr           = '0.0.0.0',
-  $port           = 5672,
-  $sasl_mech      = 'ANONYMOUS',
-  $role           = undef,
-  $idle_timeout   = undef,
+  Optional[String] $ssl_profile = undef,
+  Optional[String] $host = undef,
+  Integer[0, 65535] $port = 5672,
+  Optional[String] $sasl_mech = 'ANONYMOUS',
+  Optional[String] $role = undef,
```
ekohl: From the docs:
> role (One of [normal, inter-router, route-container], default=normal)

So can be ```Optional[Enum['normal', 'inter-router', 'route-container']]```

stbenjam:
I updated the $role to be an enum, can you take another look?

netsuro:
Hello,
Do you know when this change will be effective in repos ?
Because when i do fresh install of katello 3.5, I still do not have these changes and the qdrouterd.service service is still in error

PR #88, opened 2018-03-01
refs #21350 - allow restricting ciphers and protocols

stbenjam:


stbenjam:
@theforeman-bot is unhappy, but this issue spans many different repos.

templates/router/ssl_profile.conf.erb
```diff
@@ -6,5 +6,11 @@ ssl-profile {
 <% unless [nil, :undefined, :undef, ''].include?(@password) -%>
     password: <%= @password %>
 <% end -%>
+<% unless [nil, :undefined, :undef, ''].include?(@ciphers) -%>
+    ciphers: <%= @ciphers %>
+<% end -%>
+<% unless [nil, :undefined, :undef, ''].include?(@protocols) -%>
+    protocols: <%= @protocols.join(" ") %>
```
sean797: The current version of `qpid-dispatch-router` in epel doesn't support this parameter. I assume it is coming in a future release?

stbenjam: Yes, sorry.    It's already been merged, but waiting for an answer from the qpid devs about when we'll see it in EPEL.

ekohl: Since the default is undef, I'd be OK with including this patch.

stbenjam:
Ok thanks!

PR #95, opened 2018-07-18
Fix dependency cycle with systemd_limits and allow absent

ehelms:
This fixes two things:

 1) A dependency cycle that exists if both qpid and the router systemd limits are set at the same time
 2) Ensuring that if no limits are set or they are removed from something like a hiera config, that they are removed

```
[ERROR 2018-07-17T15:03:32 main]  Found 1 dependency cycle:
[ERROR 2018-07-17T15:03:32 main] (Concat_file[/etc/qpid-dispatch/qdrouterd.conf] => Concat[/etc/qpid-dispatch/qdrouterd.conf] => Class[Qpid::Router::Config] => Class[Qpid::Router::Service] => Systemd::Service_limits[qdrouterd.service] => File[/etc/systemd/system/qdrouterd.service.d/90-limits.conf] => Class[Systemd::Systemctl::Daemon_reload] => Exec[systemctl-daemon-reload] => Class[Systemd::Systemctl::Daemon_reload] => Exec[restart qpidd.service because limits] => Systemd::Service_limits[qpidd.service] => Service[qpidd] => Exec[migrate_pulp_db] => Class[Pulp::Database] => Class[Pulp] => Class[Foreman_proxy_content::Dispatch_router] => Qpid::Router::Ssl_profile[client] => Concat::Fragment[qdrouter+ssl_client.conf] => Concat_fragment[qdrouter+ssl_client.conf] => Concat::Fragment[qdrouter+ssl_client.conf] => Concat[/etc/qpid-dispatch/qdrouterd.conf] => Concat_file[/etc/qpid-dispatch/qdrouterd.conf])\nTry the '--graph' option and opening the resulting '.dot' file in OmniGraffle or GraphViz
```

### DIFF
--- a/manifests/router/service.pp
+++ b/manifests/router/service.pp
@@ -11,12 +11,20 @@ class qpid::router::service {
     hasrestart => true,
   }
 
-  if $::qpid::router::open_file_limit and $::systemd {
+  if $::qpid::router::open_file_limit != 1 {
+    $ensure_limit = 'present'
+  } else {
+    $ensure_limit = 'absent'
+  }
+
+  if $::systemd {
     systemd::service_limits { 'qdrouterd.service':
-      limits => {
+      ensure          => $ensure_limit,
+      restart_service => false,
+      notify          => Service['qdrouterd'],
+      limits          => {
         'LimitNOFILE' => $::qpid::router::open_file_limit,
       },
-      notify => Service['qdrouterd'],
     }
   }
 }
--- a/spec/classes/qpid_client_spec.rb
+++ b/spec/classes/qpid_client_spec.rb
@@ -4,7 +4,7 @@ describe 'qpid::client' do
   on_supported_os.each do |os, facts|
     context "on #{os}" do
       let :facts do
-        facts
+        facts.merge!(:systemd => true)
       end
 
       context 'without parameters' do
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -14,7 +14,7 @@ class qpid::params {
   $interface = undef
 
   $max_connections = undef
-  $open_file_limit = undef
+  $open_file_limit = 1 # 1 is used to denote absent condition due to how systemd::service_limits works
 
   $ssl                     = false
   $ssl_port                = 5671
--- a/manifests/router/params.pp
+++ b/manifests/router/params.pp
@@ -8,5 +8,5 @@ class qpid::router::params {
   $router_mode         = 'interior'
   $worker_threads      = $::processorcount
   $router_packages     = ['qpid-dispatch-router']
-  $open_file_limit     = undef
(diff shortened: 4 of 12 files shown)


manifests/router/service.pp
```diff
@@ -11,12 +11,18 @@
     hasrestart => true,
   }
 
-  if $::qpid::router::open_file_limit and $::systemd {
+  if $::qpid::router::open_file_limit != 1 {
```
ekohl: You can leave out the `!= 1` and just test for undef.

manifests/router/service.pp
```diff
@@ -11,12 +11,18 @@
...
+    $ensure_limit = 'absent'
+  }
+
+  if $::systemd {
     systemd::service_limits { 'qdrouterd.service':
+      ensure => $ensure_limit,
       limits => {
         'LimitNOFILE' => $::qpid::router::open_file_limit,
       },
-      notify => Service['qdrouterd'],
```
ekohl: I'm leaning to keeping this in and set `restart_service` to false. We will need to chain `Class['systemd::systemctl::daemon_reload'] -> Service['qdrouterd']` but the advantage is that we dont have a separate `systemctl restart qdrouterd` and a potential restart through the `Service`.

manifests/router/service.pp
ehelms: I cannot because if you set this to undef, the systemd:system_limits bombs out if trying to ensure absent. I would say this is a bug in the systemd:system_limit logic and this my work around.

manifests/router/service.pp
ehelms: What is the advantage of this vs. the builtin method that systemd:system_limits provides?

manifests/router/service.pp
ekohl: Ah, nasty.

manifests/router/service.pp
ekohl: The built in method can have a double restart: `exec { 'systemctl restart ...': }` + `service { '...': }`. The advantage of theirs is that you don't have to manage the service resource but if you do, the double restart is not worth it IMHO.

manifests/router/service.pp
ekohl: Maybe it's better to specify two resources, one with limits and absent and the other without?

spec/classes/qpid_client_spec.rb
```diff
@@ -4,7 +4,7 @@
   on_supported_os.each do |os, facts|
     context "on #{os}" do
       let :facts do
-        facts
+        facts.merge(:systemd => false)
```
ekohl: In which case do we run without systemd? el7 is the only supported OS. Maybe we should set a custom fact to always true in the spec helper so we don't need to override it.

ehelms:
@ekohl Updated and tests are passing, I was not able to get the top level fact setting to work reliably

PR #96, opened 2018-07-24
Properly handle systemd_limits

ekohl:
This now allows the purging of systemd limits. It also removes a duplicate restart.

Should replace https://github.com/theforeman/puppet-qpid/pull/95


ehelms:
@ekohl Mind doing a release with this fix?

PR #98, opened 2018-07-26
Fixes #24415 - Add hello interval/max_age params to qdrouterd.

chris1984:
@ekohl I am having an issue getting the tests to work correctly, when you get a chance can you provide some input please?

### DIFF
--- a/templates/router/header.conf.erb
+++ b/templates/router/header.conf.erb
@@ -24,5 +24,11 @@ router {
     id: <%= scope['qpid::router::router_id'] %>
     mode: <%= scope['qpid::router::router_mode'] %>
     worker-threads: <%= scope['qpid::router::worker_threads'] %>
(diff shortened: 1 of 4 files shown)


templates/router/header.conf.erb
```diff
@@ -24,5 +24,10 @@ router {
     id: <%= scope['qpid::router::router_id'] %>
     mode: <%= scope['qpid::router::router_mode'] %>
     worker-threads: <%= scope['qpid::router::worker_threads'] %>
+    <% unless [nil, :undefined, :undef, ''].include?(scope['qpid::router::hello_interval']) -%>
+    helloInterval: <%= scope['qpid::router::hello_interval'] %>
+    <% end %>
```
ekohl: This should be `<% end -%>` (and below as well). I believe your whitespace issue is then fixed in the tests.

chris1984:
@ekohl thanks again for the help, fixed the logic statements syntax too and all green.

PR #107, opened 2018-12-19
Allow arbitrary settings

ekohl:


evgeni:
The diff looks good to me, the test failures seem unrelated.

One question, how should we handle the case when the user passes an option that is already handled in the "named" part of the module? Currently, we just write it a second time to the config file. Will qpidd tolerate that? If it will, I think it's ok to leave it as is.

pmoravec:
If `qpidd.conf` contains an option multiple times, then `qpidd` refuses to start with explicit complain like:

    [Broker] critical Unexpected error: Error in configuration file /etc/qpid/qpidd.conf: option 'auth' cannot be specified more than once

Since the duplicated option name is named in the error log, it should be easy to identify it as a config mismatch.

evgeni:
@pmoravec thanks for testing this. I don't think we need a full config validation in the module, so either we rely on the service giving a nice error or just blacklist the values we already have as named options. I'm fine with either :)

ekohl:
Updated with a note. Since it's for advanced users only and @pmoravec's comment I think this should suffice.

ehelms:
How do I remove an option that I added this way? As a user, I would put this into my custom hiera file. Then remove an option and expect it to disappear.

ekohl:
It's not like postgresql which manages individual lines inside a file. We fully manage the file here and replace it if needed.

PR #108, opened 2018-12-19
Refactor router testing

ekohl:
This attempts to test as much of the functionality close to the file they're actually defined.


spec/defines/router_log_spec.rb
```diff
@@ -1,58 +1,97 @@
...
+          ])
+        end
 
-    it do
-      is_expected.to contain_concat('/etc/qpid-dispatch/qdrouterd.conf')
-        .that_notifies('Service[qdrouterd]')
-        .that_subscribes_to('Concat_fragment[qdrouter+log_example.conf]')
+        it do
+          is_expected.to contain_concat_fragment('qdrouter+log_example.conf')
+            .that_notifies(['Concat[/etc/qpid-dispatch/qdrouterd.conf', 'Service[qdrouterd]'])
```
evgeni: this should be `.that_notifies(['Concat[/etc/qpid-dispatch/qdrouterd.conf]', 'Service[qdrouterd]'])`

PR #113, opened 2019-01-23
Fixes #25909 - make qpidd.service wait until the port is open

evgeni:
qpidd does not open the ssl-port immediately after starting, but after
some initialization happened. Let's make qpidd.service not return until
the port is really open, as otherwise actions depending on the service
being available might fail

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -33,11 +33,9 @@ class qpid::service {
       notify  => Service['qpidd'],
     }
 
-    if (!defined(Package['nc']) and $qpid::ssl) {
-      package { 'nc':
-        ensure => installed,
-        notify => Systemd::Dropin_file['wait-for-port.conf'],
-      }
+    if $qpid::ssl {
+      ensure_packages(['nc'])
+      Package['nc'] -> Systemd::Dropin_file['wait-for-port.conf']
     }
   }
 }
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -27,6 +27,7 @@ class qpid::service {
     }
 
     systemd::dropin_file { 'wait-for-port.conf':
+      ensure  => bool2str($qpid::ssl, 'present', 'absent'),
       unit    => 'qpidd.service',
       content => template('qpid/wait-for-port.conf.erb'),
       notify  => Service['qpidd'],
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -25,5 +25,11 @@ class qpid::service {
       limits          => $limits,
       notify          => Service['qpidd'],
     }
+
+    systemd::dropin_file { 'wait-for-port.conf':
+      unit    => 'qpidd.service',
+      content => template('qpid/wait-for-port.conf.erb'),
+      notify  => Service['qpidd'],
+    }
   }
 }
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -33,7 +33,7 @@ class qpid::service {
       notify  => Service['qpidd'],
     }
 
-    if ! defined(Package['nc']) {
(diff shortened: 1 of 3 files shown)


templates/wait-for-port.conf.erb
```diff
@@ -0,0 +1,4 @@
+[Service]
+<% if scope['qpid::ssl'] -%>
```
ekohl: Could be moved to the class:
```puppet
systemd::dropin_file { 'wait-for-port.conf':
  ensure => bool2str($qpid::ssl, 'present', 'absent'),
  # ...
}
```

evgeni: good idea, updated

evgeni:
@ehelms systemd can do this for you, if you let systemd manage the socket for you, but not  if you open it yourself.

evgeni:
> Do we need to make sure nc is installed?

`libvirt-libs` depends on `nc`, and we have `foreman-libvirt → tfm-rubygem-ruby-libvirt → libvirt-libs` as a dependency chain anyways.

So theoretically, yes. But I see few chances of `nc` being not available.

ekohl:
In upstream `foreman-libvirt` isn't always installed (most of the time it isn't) so I'd prefer it to be ensured explicitly.

evgeni:
agreed and updated

manifests/service.pp
```diff
@@ -25,5 +25,19 @@
...
+      ensure  => bool2str($qpid::ssl, 'present', 'absent'),
+      unit    => 'qpidd.service',
+      content => template('qpid/wait-for-port.conf.erb'),
+      notify  => Service['qpidd'],
+    }
+
+    if ! defined(Package['nc']) {
+      package { 'nc':
+        ensure => installed,
+        notify => Systemd::Dropin_file['wait-for-port.conf'],
```
ekohl: It's better to set `require => Package['nc']` on the `dropin_file`. If it's defined in another module you won't get a correct dependency in the current implementation.

Probably easiest to use:
```puppet
if $qpid::ssl {
  ensure_packages(['nc'])
  Package['nc'] -> Systemd::Dropin_file['wait-for-port.conf']
}
```

evgeni: yepp!

templates/wait-for-port.conf.erb
```diff
@@ -0,0 +1,2 @@
+[Service]
+ExecStartPost=-/bin/bash -c 'while ! nc -z 127.0.0.1 <%= scope['qpid::ssl_port'] %>; do sleep 1; done'
```
ekohl: I do wonder about the `-` before `/bin/bash`. Why is the exit code irrelevant here? Shouldn't it eventually return true?

evgeni: And you think I still know why I added it yesterday? Yes, it will probably return 0 - or end up in an endless loop.

evgeni: Ah, I *think* I wanted to add a timeout to this, and that would produce a non-zero code. But I wanted to ignore it, just for the case where maybe we try to poke the wrong interface or something.

evgeni: So it'd be smth like:
```
ExecStartPost=-/usr/bin/timeout 60s /bin/bash >…
```

(timeout is in coreutils, no need to add a dep here :))

ekohl: Doesn't the systemd unit itself have a timeout already?

evgeni: droped the -

evgeni:
@sean797 given you had comments before, can you please have a look at the current version?

sean797:
LGTM :+1: thanks @evgeni

PR #114, opened 2019-01-30
Release 4.4.1

ekohl:
I have marked this as a patch release but I'm considering a minor version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [4.4.1](https://github.com/theforeman/puppet-qpid/tree/4.4.1) (2019-02-14)
+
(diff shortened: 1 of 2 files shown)


evgeni:
Would you explain why you would consider a minor bump?
It's fixing a bug, no new (Puppet) dependencies are introduced and no behavior change happens, so imho patch is fine.

ekohl:
That was also my reasoning but wasn't sure about potential impact. Of course your code is perfect and would never contain bugs, but I sometimes bump to a minor when a bugfix is larger. That said, I'm fine with considering this a patch release.

evgeni:
never said my code is even good ;-) (but it seems to work XD)

ehelms:
@ekohl Letting you merge this so you can release it at the same time

PR #115, opened 2019-01-31
Fixes #25909 - make qpidd.service wait until the port is open

ehelms:
qpidd does not open the ssl-port immediately after starting, but after
some initialization happened. Let's make qpidd.service not return until
the port is really open, as otherwise actions depending on the service
being available might fail

(cherry picked from commit 7534cd902129921d5f86722ea9e41df7ff188f52)

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -26,5 +26,17 @@ class qpid::service {
       limits          => $limits,
       notify          => Service['qpidd'],
     }
+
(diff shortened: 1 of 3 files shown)


evgeni:
the manifest diff looks correct, now we just need a :green_apple:

ehelms:
@evgeni It's :green_book:

PR #117, opened 2019-02-25
Move to using python2 package names

ehelms:


manifests/params.pp
```diff
@@ -32,7 +32,7 @@
   $user = 'qpidd'
   $group = 'qpidd'
 
-  $server_packages = ['qpid-cpp-server', 'qpid-cpp-client', 'python-qpid-qmf', 'python-qpid', ]
+  $server_packages = ['qpid-cpp-server', 'qpid-cpp-client', 'python2-qpid-qmf', 'python2-qpid', ]
```
ekohl: On CentOS 7 these packages don't exist.

ehelms: They do!

http://mirror.oss.ou.edu/epel/7/x86_64/Packages/p/python2-qpid-qmf-1.39.0-1.el7.x86_64.rpm
http://mirror.oss.ou.edu/epel/7/x86_64/Packages/p/python2-qpid-1.37.0-4.el7.noarch.rpm

ekohl: Ah, in EPEL. Interesting that there are much older versions in CentOS7 extras (with older versions). That threw me off.

ehelms:
Fair, that issue is worth re-visiting.

PR #124, opened 2019-05-02
Drop python2-qpid as a server package

ehelms:
This is typically needed by service interacting with qpidd and not
qpidd itself.

@ekohl you were right :)

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -35,7 +35,7 @@ class qpid::params {
   $user = 'qpidd'
   $group = 'qpidd'
 
-  $server_packages = ['qpid-cpp-server', 'qpid-cpp-client', 'python2-qpid']
+  $server_packages = ['qpid-cpp-server', 'qpid-cpp-client']
 
   $server_store = true
   $server_store_package = 'qpid-cpp-server-linearstore'


manifests/params.pp
```diff
@@ -35,7 +35,7 @@
   $user = 'qpidd'
   $group = 'qpidd'
 
-  $server_packages = ['qpid-cpp-server', 'qpid-cpp-client', 'python2-qpid']
+  $server_packages = ['qpid-cpp-server', 'qpid-cpp-client']
```
ekohl: You'd almost thing we just need qpid-cpp-server, but in this module we probably need the client package for queue actions.

ehelms:
@ekohl Mind doing a 5.0.1? Or pointing me at what instructions you use these days to release modules?

ekohl:
https://theforeman.org/plugins/katello/developers.html#HowtoReleasepuppetmodules is what I use though it's a bit of a brain dump. Since I'm on airport wifi I'll give the short instructions
```shell
PUPPET_VERSION=5.5 bundle update # 6.0 requires PDK to release and we don't have that in our gem file
bundle exec rake module:bump:minor # I'd consider this a minor release rather than a patch
git checkout -b release-$(bundle exec rake module:version)
bundle exec rake changelog # You probably need to pass a GH personal token due to API limits - it shows instructions
# Now review the changelog
git add CHANGELOG.md metadata.json
git commit -m "Release $(bundle exec rake module:version)"
```

You may want to submit a PR based on that.

Once it's merged:
```shell
bundle exec rake module:tag
git verify-tag $(bundle exec rake module:version) # probably redundant, but I always like to verify it's properly GPG signed
git push --follow-tags # push the new tag to GH
bundle exec rake module:clean module:push # assumes you have the credentials set up
```

PR #135, opened 2020-05-12
Release 7.0.0

wbclark:


ekohl:
At this point I think we should not release this module. There are no changes other than infrastructure. Modern facts are of no real value at this point since they don't even emit deprecation warnings. It's fine to let it sit in master for a while until we do need actual changes.

PR #137, opened 2020-11-11
Fixes #30252: Make wait for port more resilient

ehelms:
The current method can result in journal error messages:

[System] error Error reading socket: Encountered end of file [-5938]

This updated version waits and checks for listening ports
and greps for the port in question to appear before signaling
that the service is ready.

### DIFF
--- a/templates/wait-for-port.conf.erb
+++ b/templates/wait-for-port.conf.erb
@@ -1,2 +1,2 @@
 [Service]
-ExecStartPost=/bin/bash -c 'while ! nc -z localhost <%= scope['qpid::ssl_port'] %>; do sleep 1; done'
+ExecStartPost=/bin/bash -c 'while ! ss --no-header --tcp --listening --numeric sport = :<%= scope['qpid::ssl_port'] %> | grep -q "^LISTEN.*:<%= scope['qpid::ssl_port'] %>"; do sleep 1; done'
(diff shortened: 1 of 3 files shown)


templates/wait-for-port.conf.erb
```diff
@@ -1,2 +1,2 @@
 [Service]
-ExecStartPost=/bin/bash -c 'while ! nc -z localhost <%= scope['qpid::ssl_port'] %>; do sleep 1; done'
+ExecStartPost=/bin/bash -c 'while ! ss --no-header --tcp --listening --numeric sport = :<%= scope['qpid::ssl_port'] %> | grep -q "^LISTEN.*:<%= scope['qpid::ssl_port'] %>"; do sleep 1; done'
```
evgeni, line 2: I guess we then should also adjust the manifest to install `ss`?

https://github.com/theforeman/puppet-qpid/blob/68f944ca15e7bd0d08ec5b06b430994a2e94c5e5/manifests/service.pp#L36-L39

evgeni: (it's in `iproute`)

ehelms:
Thanks @evgeni for that point out, I have updated it and the tests.

evgeni (approved):

ekohl (approved):
This probably also solved https://projects.theforeman.org/issues/27637 or at least make it invalid.

evgeni:
Agree, it should fix that issue.

PR #138, opened 2020-11-24
Store qdrouterd connector password in separate file

wbclark:


templates/router/connector.conf.erb
```diff
@@ -7,7 +7,7 @@ connector {
     sasl-username: <%= @sasl_username %>
 <% end -%>
 <% unless [nil, :undefined, :undef, ''].include?(@sasl_password) -%>
-    sasl-password: <%= @sasl_password %>
+    sasl-password: <%= @sasl_password_file %>
```
ehelms, line 10: Should this be `sasl-password-file` now?

wbclark: I'm not sure what you mean -- are you talking about the `unless` on L9? The file resource should only get created when `defined($sasl_password)`; also `$sasl_password_file` has a default value so I think it works this way.

manifests/router/connector.pp
```diff
@@ -25,12 +27,25 @@
...
   Optional[String] $sasl_username = undef,
   Optional[String] $sasl_password = undef,
+  Stdlib::Absolutepath $sasl_password_file = '/etc/qpid-dispatch/connector_pw.conf',
   Optional[Enum['normal', 'inter-router', 'route-container']] $role = undef,
   Optional[String] $ssl_profile = undef,
   Optional[Integer[0]] $idle_timeout = undef,
   String $config_file = $qpid::router::config_file,
 ){
 
+  if defined('$sasl_password') {
```
ekohl, line 37: This should be sufficient
```suggestion
  if $sasl_password {
```

manifests/router/connector.pp
```diff
@@ -25,12 +27,25 @@
...
+    include qpid::params
+
+    file { $sasl_password_file:
```
ekohl, line 40: If there's no password, should this file be ensured absent?

manifests/router/connector.pp
```diff
@@ -25,12 +27,25 @@
...
+      ensure  => file,
+      owner   => $qpid::params::user,
+      group   => $qpid::params::group,
+      mode    => '0600',
```
ekohl, line 44: I'd suggest to make it readable, but not writable:
```suggestion
      owner   => 'root',
      group   => $qpid::params::group,
      mode    => '0640',
```

ekohl (changes requested):

templates/router/connector.conf.erb
ehelms: Well, if I look at https://qpid.apache.org/releases/qpid-dispatch-0.8.1/man/qdmanage.html I see `sasl-password-file`. If I look at https://qpid.apache.org/releases/qpid-dispatch-1.14.0/man/qdrouterd.conf.html it appears you would need to do `file:<%= @sasl_password_file %>`.

manifests/router/connector.pp
wbclark: Am I correct in understanding that we should never use `qpid::router::connector` multiple times? I don't want to create a situation where connector A creates the file with default path, and connector B has no password and clobbers it.

manifests/router/connector.pp
wbclark: Thanks. I'm never quite sure about this one.

manifests/router/connector.pp
wbclark: I think that is correct so I went ahead and implemented it.

wbclark:
Thanks for the quick reviews @ehelms @ekohl . I've pushed an update based on all feedback.

manifests/router/connector.pp
ehelms: Today, no. But possibly. I have been investigating a performance BZ around multiple connectors and links: https://bugzilla.redhat.com/show_bug.cgi?id=1754894

ekohl: It's a define, so you're right that it could be.

manifests/router/connector.pp
```diff
@@ -25,12 +27,29 @@
   Optional[String] $sasl_mech = 'ANONYMOUS',
   Optional[String] $sasl_username = undef,
   Optional[String] $sasl_password = undef,
+  Stdlib::Absolutepath $sasl_password_file = '/etc/qpid-dispatch/connector_pw.conf',
```
ekohl, line 30: How about including the title here:
```suggestion
  Stdlib::Absolutepath $sasl_password_file = '/etc/qpid-dispatch/connector-${name}-pw.conf',
```

manifests/router/connector.pp
ekohl: `defined()` checks if the variable is defined. In this case the variable was always defined but with the value `undef`. `defined()` was mostly useful back before there was `$facts` and all facts were on the top scope. That allowed you to have a code path there the fact was not present. The same with global variables, for example set via the ENC. In most cases it's not needed anymore.

manifests/router/connector.pp
ekohl: Though not 100% sure about `$name` vs `$title` here. Perhaps you have to use `$title` but the unit tests should make it quite easy to figure out.

templates/router/connector.conf.erb
wbclark: Ah, yeah. I was looking at https://qpid.apache.org/releases/qpid-dispatch-1.14.0/man/qdrouterd.conf.html but missed the `file:`. Thanks!

PR #142, opened 2020-11-30
Bump master to 7.0.0

ekohl:
There was a breaking change but also a 6.2.0 release. It's confusing to have an older version than what's actually released.


ekohl:
Merging this will also need PRs to the modules that use it to allow the newer version. It's best to prepare those so they can be merged together. Otherwise the nightly installer breaks.

PR #143, opened 2021-01-26
Release 7.0.0

ehelms:


CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+
+- Fixes [\#30252](https://projects.theforeman.org/issues/30252): Make wait for port more resilient [\#137](https://github.com/theforeman/puppet-qpid/pull/137) ([ehelms](https://github.com/ehelms))
+
+## [6.2.0](https://github.com/theforeman/puppet-qpid/tree/6.2.0) (2020-11-30)
+
+[Full Changelog](https://github.com/theforeman/puppet-qpid/compare/6.1.0...6.2.0)
+
+**Breaking changes:**
+
+- Use modern facts [\#134](https://github.com/theforeman/puppet-qpid/issues/134)
```
ekohl, line 17: This doesn't make sense and I don't think this matches reality. 6.2 should not have a breaking change.

ekohl (changes requested):

ehelms:
Does this look better?

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## [7.0.0](https://github.com/theforeman/puppet-qpid/tree/7.0.0) (2021-01-28)
+
+[Full Changelog](https://github.com/theforeman/puppet-qpid/compare/6.1.0...7.0.0)
+
+**Implemented enhancements:**
+
+- Fixes [\#30252](https://projects.theforeman.org/issues/30252): Make wait for port more resilient [\#137](https://github.com/theforeman/puppet-qpid/pull/137) ([ehelms](https://github.com/ehelms))
```
ekohl, line 9: This was already released as part of 6.2.0. Please include the changelog parts from a1a361fef88941f788e3db585debda1b43ef9d4d.

ekohl (changes requested):

ehelms:
Added the 6.2.0 section

CHANGELOG.md
ekohl: I'd drop this section since it's the same as 6.2.0. So effectively we do a major version just for the modern facts. That's why I've been holding off on a release last time: for the end user it doesn't really make a difference.

ekohl:
I released 7.0.0 now with another fix included.

PR #148, opened 2021-03-08
Refs #32037: Add parameters to disable qpid services

ehelms:


wbclark (approved):
This looks very straightforward.

I don't see a particular need to set defaults via `$qpid::{router::,}params::service_{ensure,enable}` at this time as I don't see anywhere in our ecosystem where we include either params class. (I don't recall what were the other benefits, if any, of adhering to that pattern)

I wouldn't mind seeing a test case for the proper default behavior, and another test that the defaults can be overridden, but it feels very trivial given the simplicity of this code to hold things up over it. I am assuming your reasoning for not implementing that was probably similar.

manifests/init.pp
```diff
@@ -9,7 +9,7 @@
 # $acl_file::                 File name for Qpid ACL
 #
 # $acl_content::              Content for Access Control List file
-
+#
```
ekohl, line 12: Nice additional fix that we should have caught earlier

manifests/init.pp
```diff
@@ -87,6 +91,8 @@
   Optional[Integer[1]] $mgmt_pub_interval = $qpid::params::mgmt_pub_interval,
   Optional[Integer[1]] $default_queue_limit = $qpid::params::default_queue_limit,
   Hash[String, Variant[String, Integer]] $custom_settings = $qpid::params::custom_settings,
+  Boolean $service_ensure = true,
+  Boolean $service_enable = true,
```
ekohl, line 95: I wonder if it makes sense for us to combine it. For example
```suggestion
  Boolean $service_ensure = true,
  Optional[Boolean] $service_enable = undef,
```

And later use `enable => pick($service_enable, $service_ensure)`

manifests/router.pp
```diff
@@ -16,6 +16,10 @@
 #   Frequency in seconds of sending heartbeats from this router
 # @param hello_max_age
 #   Timeout of heartbeat before connections are dropped
+# @param service_enable
+#   Enable/disable qdrouterd service at boot
+# @param service_ensure
+#   Specify if qdourterd should be running or stopped.
```
ekohl, line 22:
```suggestion
#   Specify if qdrouterd should be running or stopped.
```

ekohl (commented):
I'm ok with this if you fix the typo.

manifests/init.pp
ehelms: For our use case they could be combined, and I'll probably still have our code pass both to be explicit. So unless there is a usecase for supporting one parameter, I'd just leave it at two.

ekohl: In my suggestion you still have them both but you only need to specify one if you want them to match.

wbclark: Interesting idea. I have a few thoughts:

In the proposed usage in puppet-{foreman_proxy_content,katello} we always ensure these parameters have the same value anyway, so I don't think it makes any difference to a foreman-installer user. It would allow us to remove a line of code over there, but might also make things less readable / make the behavior a little less obvious.

So for the folks who are directly using this module, I think it comes down to: is this pattern ubiquitous enough for services that we'd be violating user expectations by not adhering to it? If not, then perhaps we could leave it for now?

manifests/router.pp
wbclark: Nice catch :)

manifests/init.pp
wbclark: I should refresh the page before posting comments :)

ekohl: I think it's hard to misuse. In most cases you want them to match anyway. If you don't, you can set them both. If both are explicit parameters, you must remember to set them both. I'd guess that in 99% cases they match and this optimizes for it.

manifests/router/service.pp
```diff
@@ -4,8 +4,8 @@
 class qpid::router::service {
 
   service { 'qdrouterd':
-    ensure     => running,
-    enable     => true,
+    ensure     => $qpid::router::service_ensure,
+    enable     => $qpid::router::service_enable,
```
wbclark, line 8: Is it intentional or an oversight that we don't use the same pattern here?

```suggestion
    enable     => pick($qpid::router::service_enable, $qpid::router::service_ensure),
```

If there is no reason for the different behavior here, then let's make it consistent.

ehelms: Blind addressing of direct comments :) I have updated both to the same pattern now with tests.

wbclark (approved):

manifests/router/service.pp
wbclark: Great, thanks!

manifests/router.pp
```diff
@@ -25,6 +29,8 @@
   $worker_threads          = $qpid::router::params::worker_threads,
   $hello_interval          = $qpid::router::params::hello_interval,
   $hello_max_age           = $qpid::router::params::hello_max_age,
+  $service_ensure          = true,
+  $service_enable          = undef,
```
ekohl, line 33: I would have liked to see data types here. Ideally whenever you add a parameter, you use a data type.

ehelms: Given this class did not have data types already, it was either add it just for these two, add it for all (and possibly overreach the point of the PR) or stick with what was present already. I chose the latter to make the focused changed. And plan to follow up adding types.

PR #149, opened 2021-03-19
Install cyrus-sasl-plain package

ehelms:
Given the SASL mechanisms used, the cyrus-sasl-plain package is
needed to support them.

This allows us to remove this requirement from the katello package spec file and localize it with the code that needs it. This further helps when we make katello-agent infrastructure optional as we will only install this package when its needed.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -9,6 +9,10 @@ class qpid::install {
     before => Class['qpid::tools'],
   }
 
+  if $qpid::auth {
+    ensure_packages(['cyrus-sasl-plain'])
+  }
+
   if $qpid::server_store {
     package { $qpid::server_store_package:
       ensure => $qpid::version,
--- a/spec/classes/qpid_spec.rb
+++ b/spec/classes/qpid_spec.rb
@@ -219,6 +219,21 @@ describe 'qpid' do
           ])
         end
       end
+
+      context 'with auth' do
+        let(:params) { super().merge(auth: true) }
+
+        it 'should create configuration file' do
+          verify_exact_contents(catalogue, '/etc/qpid/qpidd.conf', [
+            'log-enable=error+',
+            'log-to-syslog=yes',
+            'auth=yes',
+          ])
+        end
+        it 'should install cyrus-sasl-plain' do
+          is_expected.to contain_package('cyrus-sasl-plain')
+        end
+      end
     end
   end
 end


ekohl (changes requested):
Looks like c0c8e9a439b4343628e95c91d48d38c014d4b691 introduced this. Notably, in the router. That suggests that it's also needed here so it's also installed on servers with just a router:
https://github.com/theforeman/puppet-qpid/blob/6d3f06f98fae054e6c201f34845f50030c8ae785/manifests/router/params.pp#L9

Technically, it should only be needed in the qpid class if `qpid::auth` is true if I read its docs correctly.

ehelms:
> Looks like [c0c8e9a](https://github.com/theforeman/puppet-qpid/commit/c0c8e9a439b4343628e95c91d48d38c014d4b691) introduced this. Notably, in the router. That suggests that it's also needed here so it's also installed on servers with just a router:
> https://github.com/theforeman/puppet-qpid/blob/6d3f06f98fae054e6c201f34845f50030c8ae785/manifests/router/params.pp#L9

That's fair. I've moved it into a block that checks if auth is true instead.



> Looks like [c0c8e9a](https://github.com/theforeman/puppet-qpid/commit/c0c8e9a439b4343628e95c91d48d38c014d4b691) introduced this. Notably, in the router. That suggests that it's also needed here so it's also installed on servers with just a router:

I'm not sure why it sets up SASL there unless by having it on the main dispatch router, this is just required to exist even as ANONYMOUS. But the cyrus-sasl packages that get installed already handle that through dependencies:

```
[root@pipe-katello-proxy-nightly-centos7 vagrant]# yum list installed | grep sasl
cyrus-sasl.x86_64                  2.1.26-23.el7            @base               
cyrus-sasl-gssapi.x86_64           2.1.26-23.el7            @base               
cyrus-sasl-lib.x86_64              2.1.26-23.el7            @anaconda           
python-saslwrapper.x86_64          0.16-5.el7               @epel               
saslwrapper.x86_64                 0.16-5.el7               @epel     
```

manifests/install.pp
```diff
@@ -10,6 +10,13 @@
     before => [Service['qpidd'],Package['qpid-tools']],
   }
 
+  if $qpid::auth {
+    package { 'cyrus-sasl-plain':
+      ensure => present,
+      before => [Service['qpidd'],Package['qpid-tools']],
```
ekohl, line 16: Why is this needed? I see it it was already there but I really wonder why it needs to be. `Service['qpidd']` should be included in `init.pp` already. Also I don't see why this package should be installed before `qpid-tools`.

ekohl: https://github.com/theforeman/puppet-qpid/pull/150 should clean that up.

wbclark (approved):
Looks like the prior feedback was all addressed. Now the implementation is very simple. If `$qpid::auth`, install the package.

ekohl (approved):

manifests/install.pp
```diff
@@ -9,6 +9,12 @@
     before => Class['qpid::tools'],
   }
 
+  if $qpid::auth {
+    package { 'cyrus-sasl-plain':
+      ensure => present,
+    }
```
ekohl, line 15: It may be better to avoid potential duplicate declarations by using [ensure_packages](https://forge.puppet.com/modules/puppetlabs/stdlib/reference#ensure_packages):
```suggestion
    ensure_packages(['cyrus-sasl-plain'])
```

I can imagine it may be declared elsewhere too.

PR #150, opened 2021-03-25
Remove duplicate relations

ekohl:
The qpidd service relation is already set in init.pp. The direct package resource is replaced with the class that encapsulates it.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -2,18 +2,17 @@
 #
 # @api private
 class qpid::install {
-
   include qpid::tools
 
   package { $qpid::server_packages:
     ensure => $qpid::version,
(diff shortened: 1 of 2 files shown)


ehelms (approved):

wbclark:
For changelog purposes, I struggle whether to label this as a bug fix or merged PR.

If I understood correctly it shouldn't change any behaviors, but makes the code cleaner and better to work with. I went with 'Bug' for now but I'm not sure about it.

ekohl:
I don't think this is a bug. When I'm unsure, I sometimes leave them unlabeled. However, in this case I'd go with enhancement. That also usually implies a minor version bump and IMHO that makes sense. With a .z change I'd only expect bugfixes and no changes in functionality (other than fewer bugs).

PR #151, opened 2021-03-29
Release 7.1.0

wbclark:
This should get https://github.com/theforeman/puppet-qpid/pull/149 as well before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## [7.1.0](https://github.com/theforeman/puppet-qpid/tree/7.1.0) (2021-03-31)
+
+[Full Changelog](https://github.com/theforeman/puppet-qpid/compare/7.0.0...7.1.0)
+
+**Implemented enhancements:**
+
+- Add types to router.pp [\#152](https://github.com/theforeman/puppet-qpid/pull/152) ([ehelms](https://github.com/ehelms))
+- Remove duplicate relations [\#150](https://github.com/theforeman/puppet-qpid/pull/150) ([ekohl](https://github.com/ekohl))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+
+[Full Changelog](https://github.com/theforeman/puppet-qpid/compare/7.0.0...7.1.0)
+
+**Implemented enhancements:**
+
+- Refs [\#32037](https://projects.theforeman.org/issues/32037): Add parameters to disable qpid services [\#148](https://github.com/theforeman/puppet-qpid/pull/148) ([ehelms](https://github.com/ehelms))
+
+**Fixed bugs:**
+
+- Remove duplicate relations [\#150](https://github.com/theforeman/puppet-qpid/pull/150) ([ekohl](https://github.com/ekohl))
```
wbclark, line 13: I debated about whether this should be labeled as a bug or merged PR.

Thinking about it more, maybe the best approach is to skip the changelog for this one entirely, since I don't think it should be relevant for an end user.

wbclark:
> I think my only question is whether we ought to merge and include #147

given that it's backwards compatible, and this could prevent us from having to do another release later, I'm all for it.

wbclark:
Updated with #147

ehelms (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # Changelog
 
+## [7.1.0](https://github.com/theforeman/puppet-qpid/tree/7.1.0) (2021-03-30)
```
ekohl, line 3: Sadly it's 2021-03-31 already :( I really wish there was some better way for it.

ekohl (approved):

wbclark:
> Could you rebase? This is not a fast forward merge now and I prefer to only use those on releases.

Sure, it is done.

PR #152, opened 2021-03-30
Add types to router.pp

ehelms:


manifests/router.pp
```diff
@@ -21,16 +21,16 @@
...
-  $mode                    = $qpid::router::params::router_mode,
-  $config_file             = $qpid::router::params::config_file,
-  $router_packages         = $qpid::router::params::router_packages,
-  $open_file_limit         = $qpid::router::params::open_file_limit,
-  $worker_threads          = $qpid::router::params::worker_threads,
-  $hello_interval          = $qpid::router::params::hello_interval,
-  $hello_max_age           = $qpid::router::params::hello_max_age,
-  $service_ensure          = true,
-  $service_enable          = undef,
+  Stdlib::Host $router_id = $qpid::router::params::router_id,
```
ekohl, line 24: I don't know how this is limited. Can it be a FQDN or an IP or just any string?

templates/router/header.conf.erb
```diff
@@ -22,7 +22,7 @@
 
 router {
     id: <%= scope['qpid::router::router_id'] %>
-    mode: <%= scope['qpid::router::router_mode'] %>
+    mode: <%= scope['qpid::router::mode'] %>
```
ekohl, line 25: This looks like an unrelated bug fix. For the changelog it may be good to have it separate. In case we want to cherry pick it.

manifests/router.pp
ehelms: > The unique identifier of the router. This ID is used as the container-id (container name) at the AMQP protocol level. If it is not specified, the router shall generate a random identifier at startup.

That reads like it could be any ole string.

templates/router/header.conf.erb
ehelms: I can extract it. This change unearthed the bug since `router_mode` got removed from params.pp this removed the parameter and tests pointed out the mismatch.

ehelms: https://github.com/theforeman/puppet-qpid/pull/153

ehelms:
Will need a rebase on top of https://github.com/theforeman/puppet-qpid/pull/153

PR #47, opened 2016-11-06
Package conflicts during install on CentOS7, fixes #46

braddeicide:
qpid-tools installs a compat package that blocks client and server if theyre not installed first, fixes #46

Issue #46, opened 2016-11-06
Installation fails without manual intervention on CentOS7


braddeicide:
qpid-tools installs compat-qpid-cpp-client which then conflicts/blocks qpid-cpp-server and qpid-cpp-client.

Error: Execution of '/bin/yum -d 0 -e 0 -y install qpid-cpp-server' returned 1: Error: compat-qpid-cpp-client conflicts with qpid-cpp-client-1.35.0-1.el7.x86_64
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Error: /Stage[main]/Qpid::Install/Package[qpid-cpp-server]/ensure: change from purged to present failed: Execution of '/bin/yum -d 0 -e 0 -y install qpid-cpp-server' returned 1: Error: compat-qpid-cpp-client conflicts with qpid-cpp-client-1.35.0-1.el7.x86_64
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Error: Execution of '/bin/yum -d 0 -e 0 -y install qpid-cpp-client' returned 1: Error: compat-qpid-cpp-client conflicts with qpid-cpp-client-1.35.0-1.el7.x86_64
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Error: /Stage[main]/Qpid::Install/Package[qpid-cpp-client]/ensure: change from purged to present failed: Execution of '/bin/yum -d 0 -e 0 -y install qpid-cpp-client' returned 1: Error: compat-qpid-cpp-client conflicts with qpid-cpp-client-1.35.0-1.el7.x86_64
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Error: Execution of '/bin/yum -d 0 -e 0 -y install qpid-cpp-server-linearstore' returned 1: Error: compat-qpid-cpp-client conflicts with qpid-cpp-client-1.35.0-1.el7.x86_64
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Error: /Stage[main]/Qpid::Install/Package[qpid-cpp-server-linearstore]/ensure: change from purged to present failed: Execution of '/bin/yum -d 0 -e 0 -y install qpid-cpp-server-linearstore' returned 1: Error: compat-qpid-cpp-client conflicts with qpid-cpp-client-1.35.0-1.el7.x86_64
[...]

PR #154, opened 2021-04-01
Refs #32037: Add ensure on qpid class to allow removal of all qpid as…

ehelms:
…sets

Adds an ensure parameter at the top level to allow specifyin 'absent'
to remove all packages, configuration files and disable services.

I have only done the qpidd part so far (will do qpid-dispatch after discussion) to allow for discussion of design and how far to take this. I have added some basic acceptance tests which will fail and illuminate some of the challenges with doing absent on the package side of things when dependencies are involved.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -4,24 +4,24 @@
 class qpid::config
 {
 
-  group { $qpid::group:
-    ensure => present,
-  }
-
   user { $qpid::user:
-    ensure => present,
+    ensure => $qpid::ensure,
     groups => $qpid::user_groups,
   }
 
+  group { $qpid::group:
+    ensure => $qpid::ensure,
+  }
+
   file { $qpid::config_file:
-    ensure  => file,
+    ensure  => $qpid::ensure,
     content => template('qpid/qpidd.conf.erb'),
     owner   => 'root',
     group   => 'root',
     mode    => '0644',
   }
 
-  if $qpid::acl_content {
+  if $qpid::acl_content and $qpid::ensure == 'present' {
     $acl_file_ensure = file
   } else {
     $acl_file_ensure = absent
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -65,6 +65,8 @@
 #
 # $service_enable::           Enable qpidd service at boot
 #
+# $ensure::                   Specify to explicitly enable Qpid installs or absent to remove all packages and configs
+#
 class qpid (
   String $version = $qpid::params::version,
   Boolean $auth = $qpid::params::auth,
@@ -93,6 +95,7 @@ class qpid (
   Hash[String, Variant[String, Integer]] $custom_settings = $qpid::params::custom_settings,
   Boolean $service_ensure = true,
   Optional[Boolean] $service_enable = undef,
+  Enum['present', 'absent'] $ensure = 'present',
 ) inherits qpid::params {
   if $ssl {
     assert_type(Boolean, $ssl_require_client_auth)
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -2,21 +2,24 @@
 #
 # @api private
 class qpid::install {
-  include qpid::tools
+
+  if $qpid::ensure == 'absent' {
+    $_package_ensure = 'purged'
+  } else {
+    $_package_ensure = $qpid::version
+  }
 
   package { $qpid::server_packages:
-    ensure => $qpid::version,
-    before => Class['qpid::tools'],
+    ensure => $_package_ensure,
   }
 
   if $qpid::auth {
-    ensure_packages(['cyrus-sasl-plain'])
+    ensure_packages(['cyrus-sasl-plain'], {ensure => $_package_ensure})
   }
 
   if $qpid::server_store {
     package { $qpid::server_store_package:
-      ensure => $qpid::version,
-      before => Class['qpid::tools'],
+      ensure => $_package_ensure,
     }
   }
 }
--- a/manifests/router/service.pp
+++ b/manifests/router/service.pp
@@ -3,15 +3,23 @@
 # @api private
 class qpid::router::service {
 
+  if $qpid::router::ensure == 'absent' {
+    $_service_ensure = false
+    $_service_enable = false
+  } else {
+    $_service_ensure = $qpid::router::service_ensure
+    $_service_enable = $qpid::router::service_enable
+  }
+
   service { 'qdrouterd':
-    ensure     => $qpid::router::service_ensure,
-    enable     => pick($qpid::router::service_enable, $qpid::router::service_ensure),
+    ensure     => $_service_ensure,
+    enable     => pick($_service_enable, $_service_ensure),
     hasstatus  => true,
     hasrestart => true,
   }
 
   if $facts['systemd'] {
-    if $qpid::router::open_file_limit {
+    if $qpid::router::open_file_limit and $qpid::router::ensure {
       $ensure_limit = 'present'
       $limits = {'LimitNOFILE' => $qpid::router::open_file_limit}
     } else {
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -3,15 +3,23 @@
 # @api private
 class qpid::service {
 
+  if $qpid::ensure == 'absent' {
+    $_service_ensure = false
+    $_service_enable = false
+  } else {
+    $_service_ensure = $qpid::service_ensure
+    $_service_enable = $qpid::service_enable
+  }
+
   service { 'qpidd':
-    ensure     => $qpid::service_ensure,
-    enable     => pick($qpid::service_enable, $qpid::service_ensure),
+    ensure     => $_service_ensure,
+    enable     => pick($_service_enable, $_service_ensure),
     hasstatus  => true,
     hasrestart => true,
   }
 
   if $facts['systemd'] {
-    if $qpid::open_file_limit {
+    if $qpid::open_file_limit and $qpid::ensure == 'present' {
       $ensure_limit = 'present'
       $limits = {'LimitNOFILE' => $qpid::open_file_limit}
     } else {
@@ -26,15 +34,21 @@ class qpid::service {
       notify          => Service['qpidd'],
     }
 
+    if $qpid::ssl and $qpid::ensure == 'present' {
+      $_ssl_ensure = 'present'
+    } else {
+      $_ssl_ensure = 'absent'
+    }
+
     systemd::dropin_file { 'wait-for-port.conf':
-      ensure  => bool2str($qpid::ssl, 'present', 'absent'),
+      ensure  => $_ssl_ensure,
       unit    => 'qpidd.service',
       content => template('qpid/wait-for-port.conf.erb'),
       notify  => Service['qpidd'],
     }
 
     if $qpid::ssl {
-      ensure_packages(['iproute'])
+      ensure_packages(['iproute'], {ensure => $qpid::ensure})
       Package['iproute'] -> Systemd::Dropin_file['wait-for-port.conf']
     }
   }
--- a/spec/acceptance/qpid_tool_spec.rb
+++ b/spec/acceptance/qpid_tool_spec.rb
@@ -14,4 +14,28 @@ describe 'qpid::tools' do
       it { is_expected.to be_installed }
     end
   end
+
+  context 'installing qpid' do
+    let(:pp) do
+      <<-PUPPET
+      class { 'qpid::tools': } ->
+      class { 'qpid': }
+      PUPPET
+    end
+
+    it_behaves_like 'a idempotent resource'
+
+    describe package('qpid-tools') do
+      it { is_expected.to be_installed }
+    end
+
+    describe service('qpidd') do
+      it { is_expected.to be_running }
+      it { is_expected.to be_enabled }
+    end
+
+    describe port('5672') do
+      it { is_expected.to be_listening }
+    end
+  end
 end
--- a/manifests/router.pp
+++ b/manifests/router.pp
@@ -20,6 +20,9 @@
 #   Enable/disable qdrouterd service at boot
 # @param service_ensure
 #   Specify if qdrouterd should be running or stopped.
+# @param ensure
+#   Specify to explicitly enable Qpid Dispatch installs or absent to remove all packages and configs
+#
 class qpid::router(
   String $router_id = $qpid::router::params::router_id,
   String $mode = 'interior',
@@ -31,6 +34,7 @@ class qpid::router(
   Optional[Integer] $hello_max_age = undef,
   Boolean $service_ensure = true,
   Optional[Boolean] $service_enable = undef,
+  Enum['present', 'absent'] $ensure = 'present',
 ) inherits qpid::router::params {
 
   contain qpid::router::install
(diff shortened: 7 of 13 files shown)


manifests/init.pp
```diff
@@ -93,6 +95,7 @@
   Hash[String, Variant[String, Integer]] $custom_settings = $qpid::params::custom_settings,
   Boolean $service_ensure = true,
   Optional[Boolean] $service_enable = undef,
+  Enum["present", "absent"] $ensure = "present",
```
wbclark, line 98:
```suggestion
  Enum['present', 'absent'] $ensure = 'present',
```

This made the linter happy for me

wbclark (commented):
When I run the acceptance test on EL7, with `ensure => absent` I get

```
centos7-64-1 23:54:25$ puppet apply --verbose --detailed-exitcodes /tmp/apply_manifest.pp.4KNmBx
  Info: Loading facts
  Info: Loading facts
  Notice: Compiled catalog for centos7-64-1 in environment production in 0.14 seconds
  Info: Applying configuration version '1617335668'
  Error: Execution of '/usr/bin/rpm -e qpid-cpp-server' returned 1: error: Failed dependencies:
  	libqpidbroker.so.2()(64bit) is needed by (installed) qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64
  	qpid-cpp-server = 1.39.0-1.el7 is needed by (installed) qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64
  Error: /Stage[main]/Qpid::Install/Package[qpid-cpp-server]/ensure: change from '1.39.0-1.el7' to 'absent' failed: Execution of '/usr/bin/rpm -e qpid-cpp-server' returned 1: error: Failed dependencies:
  	libqpidbroker.so.2()(64bit) is needed by (installed) qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64
  	qpid-cpp-server = 1.39.0-1.el7 is needed by (installed) qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64
  Error: Execution of '/usr/bin/rpm -e qpid-cpp-client' returned 1: error: Failed dependencies:
  	libqpidclient.so.2()(64bit) is needed by (installed) qpid-qmf-1.39.0-1.el7.x86_64
  	libqpidcommon.so.2()(64bit) is needed by (installed) qpid-cpp-server-1.39.0-1.el7.x86_64
  	libqpidcommon.so.2()(64bit) is needed by (installed) qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64
  	libqpidcommon.so.2()(64bit) is needed by (installed) qpid-qmf-1.39.0-1.el7.x86_64
  	libqpidmessaging.so.2()(64bit) is needed by (installed) qpid-qmf-1.39.0-1.el7.x86_64
  	libqpidtypes.so.1()(64bit) is needed by (installed) qpid-cpp-server-1.39.0-1.el7.x86_64
  	libqpidtypes.so.1()(64bit) is needed by (installed) qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64
  	libqpidtypes.so.1()(64bit) is needed by (installed) qpid-qmf-1.39.0-1.el7.x86_64
  	qpid-cpp-client = 1.39.0-1.el7 is needed by (installed) qpid-cpp-server-1.39.0-1.el7.x86_64
  	qpid-cpp-client = 1.39.0-1.el7 is needed by (installed) qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64
  	qpid-cpp-client(x86-64) = 1.39.0-1.el7 is needed by (installed) qpid-qmf-1.39.0-1.el7.x86_64
  	qpid-cpp-client(x86-64) = 1.39.0-1.el7 is needed by (installed) python2-qpid-qmf-1.39.0-1.el7.x86_64
  Error: /Stage[main]/Qpid::Install/Package[qpid-cpp-client]/ensure: change from '1.39.0-1.el7' to 'absent' failed: Execution of '/usr/bin/rpm -e qpid-cpp-client' returned 1: error: Failed dependencies:
  	libqpidclient.so.2()(64bit) is needed by (installed) qpid-qmf-1.39.0-1.el7.x86_64
  	libqpidcommon.so.2()(64bit) is needed by (installed) qpid-cpp-server-1.39.0-1.el7.x86_64
  	libqpidcommon.so.2()(64bit) is needed by (installed) qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64
  	libqpidcommon.so.2()(64bit) is needed by (installed) qpid-qmf-1.39.0-1.el7.x86_64
  	libqpidmessaging.so.2()(64bit) is needed by (installed) qpid-qmf-1.39.0-1.el7.x86_64
  	libqpidtypes.so.1()(64bit) is needed by (installed) qpid-cpp-server-1.39.0-1.el7.x86_64
  	libqpidtypes.so.1()(64bit) is needed by (installed) qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64
  	libqpidtypes.so.1()(64bit) is needed by (installed) qpid-qmf-1.39.0-1.el7.x86_64
  	qpid-cpp-client = 1.39.0-1.el7 is needed by (installed) qpid-cpp-server-1.39.0-1.el7.x86_64
  	qpid-cpp-client = 1.39.0-1.el7 is needed by (installed) qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64
  	qpid-cpp-client(x86-64) = 1.39.0-1.el7 is needed by (installed) qpid-qmf-1.39.0-1.el7.x86_64
  	qpid-cpp-client(x86-64) = 1.39.0-1.el7 is needed by (installed) python2-qpid-qmf-1.39.0-1.el7.x86_64
  Notice: /Stage[main]/Qpid::Install/Package[qpid-cpp-server-linearstore]/ensure: removed
  Info: Class[Qpid::Install]: Unscheduling all events on Class[Qpid::Install]
```

To me this looks like we are failing to remove the packages due to RPM dependencies, because it is trying to use the RPM package provider. I believe we could specify yum package provider instead which would handle this. I will give it a shot.

manifests/install.pp
```diff
@@ -4,18 +4,24 @@
 class qpid::install {
   include qpid::tools
 
+  if $qpid::ensure == 'absent' {
+    $_package_ensure = $qpid::ensure
```
wbclark, line 8:
```suggestion
    $_package_ensure = 'purged'
```

It turns out the yum provider uses `rpm -e` when `ensure => absent`, which explains the behavior I saw in the acceptance test and why my first idea did not resolve it. 

However `ensure => purged` resolved the problem by actually using yum to remove the packages. Afterwards on the test system I verified that the dependencies were correctly removed:

```
[root@centos7-64-1 ~]# yum history info 9
Loaded plugins: fastestmirror
Transaction ID : 9
Begin time     : Fri Apr  2 04:29:43 2021
Begin rpmdb    : 343:da3c175c120c70db164a3ee5d245a119f472f947
End time       :            04:29:44 2021 (1 seconds)
End rpmdb      : 339:2d5a594ccf2ee058ba67b46b501e4bdc44fab71a
User           : root <root>
Return-Code    : Success
Command Line   : -y erase qpid-cpp-client
Transaction performed with:
    Installed     rpm-4.11.3-43.el7.x86_64                      @anaconda
    Installed     yum-3.4.3-167.el7.centos.noarch               @anaconda
    Installed     yum-plugin-fastestmirror-1.1.31-53.el7.noarch @anaconda
Packages Altered:
    Erase python2-qpid-qmf-1.39.0-1.el7.x86_64 @epel
    Erase qpid-cpp-client-1.39.0-1.el7.x86_64  @epel
    Erase qpid-qmf-1.39.0-1.el7.x86_64         @epel
    Erase qpid-tools-1.39.0-1.el7.noarch       @epel
history info
[root@centos7-64-1 ~]# yum history info 8
Loaded plugins: fastestmirror
Transaction ID : 8
Begin time     : Fri Apr  2 04:29:42 2021
Begin rpmdb    : 345:b800c4ff8d099fe7d7b6f1b9715103f483b07f0d
End time       :            04:29:43 2021 (1 seconds)
End rpmdb      : 343:da3c175c120c70db164a3ee5d245a119f472f947
User           : root <root>
Return-Code    : Success
Command Line   : -y erase qpid-cpp-server
Transaction performed with:
    Installed     rpm-4.11.3-43.el7.x86_64                      @anaconda
    Installed     yum-3.4.3-167.el7.centos.noarch               @anaconda
    Installed     yum-plugin-fastestmirror-1.1.31-53.el7.noarch @anaconda
Packages Altered:
    Erase qpid-cpp-server-1.39.0-1.el7.x86_64             @epel
    Erase qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64 @epel
Scriptlet output:
   1 warning: /etc/qpid/qpidd.conf saved as /etc/qpid/qpidd.conf.rpmsave
history info
```

manifests/config.pp
```diff
@@ -5,23 +5,23 @@
 {
 
   group { $qpid::group:
-    ensure => present,
+    ensure => $qpid::ensure,
   }
 
   user { $qpid::user:
```
wbclark, line 11: Since no ordering is defined here between the user and group, Puppet will use the manifest ordering by default.

That means when running with `ensure => absent`, deleting the group will fail with: `groupdel: cannot remove the primary group of user 'qpidd'`

Changing the order in which these two resources are declared in the manifest resolved this issue for me.

manifests/install.pp
```diff
@@ -2,21 +2,24 @@
 #
 # @api private
 class qpid::install {
-  include qpid::tools
```
ehelms, line 5: I dropped this (and the before later on) as this should only be needed when running a qpid-config command which is handled by the config_cmd. If this is kept in, idempotency is broken when `ensure => absent` is set. We would then need to define the class here, and pass in a dedicated ensure to the `qpid::tools` class additionally. This does mean the qpid-tools package won't be removed but given that is a client tool I'm not too worried about it when wanting to ensure qpid the service is absent.

If a reviewer feels we should be also removing that package I'll add a parameter to the `qpid::tools` class and pass it in here.

manifests/install.pp
```diff
@@ -2,21 +2,24 @@
...
+
+  if $qpid::ensure == 'absent' {
+    $_package_ensure = 'purged'
+  } else {
+    $_package_ensure = $qpid::version
+  }
 
   package { $qpid::server_packages:
-    ensure => $qpid::version,
-    before => Class['qpid::tools'],
```
wbclark, line 9: This could in theory cause a regression of https://github.com/theforeman/puppet-qpid/pull/47 ?

manifests/install.pp
wbclark: The work involved with removing it seems small enough that I am in favor of leaving no package behind when we clean up.

manifests/install.pp
ehelms: Good question, I'll add a test to see.

spec/acceptance/qpid_tool_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper_acceptance'
+
+describe 'qpid::tools' do
```
ehelms, line 3: @wbclark Take a close look, this test appears to support that the issue in #46 is no longer an issue.

wbclark: Thanks for implementing this :+1:

wbclark (approved):
ACK, it looks good

I still would like to see the qpid-tools package removed as well for the sake of tidiness, but I don't see a need to hold up this PR over it. I'm happy to implement that myself in a follow up PR even.

wbclark:
With the following PRs now merged:

https://github.com/theforeman/puppet-qpid/pull/156
https://github.com/theforeman/puppet-qpid/pull/155

This one now needs a rebase when you have some free cycles for it.

wbclark (approved):

wbclark:
doh, there was a small typo in the full commit message that I didn't catch

or maybe you meant to emphasize a southern dialect :)

manifests/config.pp
```diff
@@ -4,24 +4,24 @@
...
-
   user { $qpid::user:
-    ensure => present,
+    ensure => $qpid::ensure,
     groups => $qpid::user_groups,
   }
 
+  group { $qpid::group:
+    ensure => $qpid::ensure,
+  }
```
ekohl, line 14: FYI: I'm not a fan of moving this around. If you need to move it, it implies a dependency. Relying on the order in the manifest is not the right way to achieve that.

manifests/install.pp
```diff
@@ -2,21 +2,24 @@
...
+    ensure => $_package_ensure,
   }
 
   if $qpid::auth {
-    ensure_packages(['cyrus-sasl-plain'])
+    ensure_packages(['cyrus-sasl-plain'], {ensure => $_package_ensure})
```
ekohl, line 17: I'm not sure if this can just be removed. It may be used by others. I'd consider changing the `if` above it to not manage it if it should be absent.

manifests/router/service.pp
```diff
@@ -3,15 +3,23 @@
...
-    enable     => pick($qpid::router::service_enable, $qpid::router::service_ensure),
+    ensure     => $_service_ensure,
+    enable     => pick($_service_enable, $_service_ensure),
     hasstatus  => true,
     hasrestart => true,
   }
 
   if $facts['systemd'] {
-    if $qpid::router::open_file_limit {
+    if $qpid::router::open_file_limit and $qpid::router::ensure {
```
ekohl, line 22: Any string in Puppet evaluates to true so it should have also checked for an equality to `present`.

manifests/service.pp
```diff
@@ -26,15 +34,21 @@
...
-      ensure  => bool2str($qpid::ssl, 'present', 'absent'),
+      ensure  => $_ssl_ensure,
       unit    => 'qpidd.service',
       content => template('qpid/wait-for-port.conf.erb'),
       notify  => Service['qpidd'],
     }
 
     if $qpid::ssl {
-      ensure_packages(['iproute'])
+      ensure_packages(['iproute'], {ensure => $qpid::ensure})
```
ekohl, line 51: This will cause Bad Things. You really don't want to remove `iproute`.

manifests/install.pp
ehelms: I dont know of anything else that should be using this on a system with qpid and specifically in our standard use case for sure; however, I could go either way with how much we do or do not cleanup. Also, I don't understand "changing the if above it to not manage it" statement.

ekohl: I'd write:
```puppet
if $qpid::auth && $qpid::ensure == 'present' {
  ensure_pacakges(['cyrus-sasl-plain'])
}
```

Essentially, when you use `ensure_packages` you say that you need the package, but another module may also need it. This is commonly used for tooling. `cyrus-sasl`-plain` may be used in other places. I recall having set up a postfix server using SASL in the past. Those kind of setups are the ones you could break.

ehelms: Ok. Educate me if you will, why would you put your postfix server on the same box as your Qpid - Katello install?

ekohl: It was just an example. Postfix could also be installed as an email relay. Foreman by default uses `/usr/bin/sendmail` which is provided by Postfix on EL. I can imagine that some might harden it where SASL is involved. For example as a relay for isolated Smart Proxies that still need to send cron reports.

And you're right that in environments where the OS is free of cost you might deploy a separate machine for better isolation but if I was a RHEL customer and would need to pay money for it, I'd probably combine that functionality.

My main point is that it's about learning the right patterns. Using `ensure_packages` with ensuring absent is a hint you may be doing it wrong. There are probably exceptions, but it should always flag a reviewer because in most cases it's using a pattern the wrong way.

PR #156, opened 2021-04-05
Add basic set of acceptance tests

ehelms:


wbclark (commented):
This is missing `spec/setup_acceptance_node.pp` which is causing the package installations to fail in acceptance tests due to repositories not being enabled.

wbclark (approved):
LGTM. Thanks!

PR #157, opened 2021-04-05
Use correct powertools repository in acceptance tests on older Centos8

wbclark:
When I run the EL8 acceptance tests locally, `qpid-cpp-server-linearstore` will fail to install due to missing dependency on `libdb-cxx`. This is because my environment defaults to an older Centos8 box where the name of the repository had not yet changed.

### DIFF
--- a/spec/setup_acceptance_node.pp
+++ b/spec/setup_acceptance_node.pp
@@ -3,7 +3,13 @@ if $facts['os']['release']['major'] == '8' {
     ensure => installed,
   }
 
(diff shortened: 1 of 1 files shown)


ehelms:
@wbclark can you pull a newer image and see if that fixes the problem? If so, I'd vote to close this as we should support the latest image available.

wbclark:
What finally worked for me to update the box is

`vagrant global-status` to get the id of the box, then
`vagrant update $id`

Closing this

PR #158, opened 2021-04-05
Refs #32037: Add ensure on qpid::tools class to allow removal of package

wbclark:


manifests/install.pp
```diff
@@ -5,8 +5,14 @@
 
   if $qpid::ensure == 'absent' {
     $_package_ensure = 'purged'
+    $_tools_ensure   = $qpid::ensure
```
ehelms, line 8: Would we not want to use purge here too?

wbclark: I felt it was more consistent to have `class qpid::tools( Enum['present', 'absent'] $ensure = 'present', )` similar to the other clases

wbclark:
I figured the test of

```
      <<-PUPPET
      class { 'qpid::tools': } ->
      class { 'qpid': }
      PUPPET
```

in spec/acceptance/qpid_tool_spec.rb would cause an issue. Working on a way to solve that while ensuring `qpid::config_cmd` still installs the tools package when needed

manifests/install.pp
```diff
@@ -5,8 +5,14 @@
...
   } else {
     $_package_ensure = $qpid::version
+    $_tools_ensure   = 'present'
+  }
+
+  class { 'qpid::tools':
+    ensure => $_tools_ensure,
```
ehelms, line 15: I think you could just pass `$qpid::ensure` directly here.

wbclark: :+1:

wbclark:
All tests are working for me locally now and I have rebased to solve the merge conflict

spec/acceptance/qpid_tool_spec.rb
```diff
@@ -15,11 +15,28 @@
...
+    end
+  end
+
   context 'installing qpid' do
     let(:pp) do
       <<-PUPPET
-      class { 'qpid::tools': } ->
-      class { 'qpid': }
+      class { 'qpid':
+        require => Class['qpid::tools'],
```
ekohl, line 38: Why is this require needed? Looking at it, it feels like `qpid::config_cmd` should have this relationship. You can also use `contain qpid::tools` in `qpid::install` to ensure the relationship. Both is also an option.

manifests/tools.pp
```diff
@@ -1,8 +1,20 @@
...
+#
+class qpid::tools(
+  Enum['present', 'absent'] $ensure = 'present',
+) {
+
+  if $qpid::tools::ensure == 'absent' {
+    $_package_ensure = 'purged'
+  } else {
+    $_package_ensure = $qpid::tools::ensure
+  }
```
ekohl, line 14: It should be in the local scope. Doesn't this also work?
```suggestion
  if $ensure == 'absent' {
    $_package_ensure = 'purged'
  } else {
    $_package_ensure = $ensure
  }
```

You can also use
```suggestion
  $_package_ensure = bool2str($ensure == 'absent', 'purged', $ensure)
```

That might also be short enough to use directly with an additional variable.

spec/acceptance/qpid_tool_spec.rb
ehelms: Note that this is in the acceptance tests. The idea with the test is to ensure that #46 is no longer a problem by first installing the qpid-tools package and then installing the qpid packages.

ekohl: But what are you testing? That if you manually provide a workaround that it doesn't show up? If you provide the correct chaining (the current patch does **not** do that), the catalog should fail to compile if you first try to install tools before the client, right?

ehelms: I am missing what is being seen as the manual work around.

The issue posed in #46 was: if you install qpid-tools before installing qpid-cpp-server, the qpid-tools package will install a 'compat' package that then blocks the installation of qpid-cpp-server. This is no longer true. You can currently install them in whatever order you want awithout one breaking the other.

The idea behind the test was to ensure that qpid-tools can be installed first, and that qpid-cpp-server can be installed afterward without running into not being able to install qpid-cpp-server due to the previous, no longer existent, compat package. To prove that we no longer needed this relationship:

https://github.com/theforeman/puppet-qpid/commit/734b44b8dd2061743f0b12695482ac691682deb5#diff-29fdc8c170b6278c50ae7a6359737e386dc4769ce94b9a6e505f5bcd3e4890deL9

manifests/tools.pp
wbclark: Thanks. I'm not sure whether the saved space in the latter example is worth the cost to readability? I'll make an update to keep `if ... { ... } else { ... }` but use local scope as in the former example.

If you feel strongly that it's better code quality to use `bool2str(...)` instead, I think there are other places where we should make that change as well.

spec/acceptance/qpid_tool_spec.rb
wbclark: I pushed a 2nd commit which adds a comment above the test case, explaining the reason for the `require => Class['qpid::tools'],` in the test case and linking to https://github.com/theforeman/puppet-qpid/issues/46

wbclark:
Thanks for the feedback @ekohl ! I've pushed updates addressing both issues and will mark the issues as resolved. If you feel either one merits further discussion, please feel free to re-open.

spec/acceptance/qpid_tool_spec.rb
ekohl: > The issue posed in #46 was: if you install qpid-tools before installing qpid-cpp-server, the qpid-tools package will install a 'compat' package that then blocks the installation of qpid-cpp-server. This is no longer true. You can currently install them in whatever order you want awithout one breaking the other.

Can you tell me how you know it's fixed? Is there no chance of an older repository still having the broken versions? I'm mostly asking because it is in EPEL7 and we have less control over that.

That's also something I really dislike about combining things in a commit. I assumed it was a regression in 734b44b8dd2061743f0b12695482ac691682deb5 because there was nothing mentioned about it at all. However, it is a big change that deserved a PR on its own. That would have informed me that it was intentional and not a side effect. My instinct now was to write patch to revert part of it.

ehelms: > Can you tell me how you know it's fixed? Is there no chance of an older repository still having the broken versions? I'm mostly asking because it is in EPEL7 and we have less control over that.

Sure an old repository could have some old version of this package and break. But that shouldn't be a supported case. We test and expect the current versions in EPEL.

In my original change, my intent was to allow the ensuring of qpid the server. As #46 is no longer a problem, and I did not want to combine ensuring qpid-tools and ensuring qpid-server it made sense to me to break the dependency (again assuming #46 was no longer an issue which it wasnt). Could I have tried to do this as separate changes? Yes. Did it occur to us at the time? No. The question now is how do we move forward with this change. Are there changes requested here?

ekohl: Then I would suggest to be explicit in the test:
```puppet
include qpid
# Old versions of qpid-tools installed compat-qpid-cpp-client causing a conflict
# https://github.com/theforeman/puppet-qpid/issues/46
# This verifies it does work now
Package['qpid-tools'] -> Package['qpid-cpp-server', 'qpid-cpp-client']
```

However, to really ensure it's a clean system then you also ensure the server packages are not present. Otherwise you're not really reproducing what was a problem.

You could also argue that its more important to test that the default invocation without any additional code is more important. We can't reproduce the problematic situation anyway (since EPEL no longer ships broken packages) so what the current code is really trying to achieve is some regression testing.

I think I'd drop this test altogether since I don't think it's really testing anything anymore.

wbclark: I've opened https://github.com/theforeman/puppet-qpid/pull/160 to drop the test. If that is satisfactory then I'll rebase this one.

manifests/tools.pp
```diff
@@ -1,8 +1,20 @@
...
+#  Specify 'present' to install qpid-tools package or 'absent' to remove the package and any dependencies
+#
+class qpid::tools(
+  Enum['present', 'absent'] $ensure = 'present',
+) {
+
+  if $ensure == 'absent' {
+    $_package_ensure = 'purged'
+  } else {
+    $_package_ensure = $ensure
```
ehelms, line 13: Should this get set to `installed` instead of present? I'm not clear from the documentation on the package type what the difference is between those two.

wbclark: AFAICT, the only difference is style. Older versions of the docs explicitly state that these are equivalent; for some reason the newer docs don't say anything about it.

Other Resource types also seem have have some allowed values for `ensure` that are aliases. I think `file { $filename: ensure => {'file','present'}` are equivalent, but it doesn't make sense to have any generic alias of `file { $path: ensure => 'directory' }`.

The Puppet 7 style guide at https://puppet.com/docs/puppet/7.5/style_guide.html has an implicit convention of using  `package { $name: ensure => 'present' }` throughout, without specifically calling that out as something one should do.

wbclark: Related style thought:

I doubt it would make much difference here, but as a general practice, is it actually better to specifically pass the string 'present' rather than the value of `$ensure`? I suppose it is easier to read that way, and allows a bit more separation of the interface for `qpid::tools` vs. the values it passes to the package resource.

wbclark: Another point is, `ensure => 'installed'` is a little more readable/greppable than `ensure => present`. I really don't have a strong preference either way. If you do, let me know and I'll make the quick change.

Otherwise, is there anything else you or @ekohl would like to review further with this PR or should we move forward?

ehelms (approved):

manifests/tools.pp
ekohl: I've always understood `installed` was just easier to read. A quick look at the source tells me they're really an alias as @wbclark said.

https://github.com/puppetlabs/puppet/blob/b9240c317780df0359eb2451db0a4cb0177a353a/lib/puppet/type/package.rb#L113-L114

Which makes sense: installed means it's present and present means it's installed. If we compare it to file type, there isn't a symmetrical relationship. You can say that directory implies it's present but just because it's present doesn't mean it's a directory. That's why it can be good to avoid `present` on the file type: it's ambiguous. That's why https://github.com/voxpupuli/puppet-lint-file_ensure-check exists.

PR #159, opened 2021-04-06
Refs #32037: Do not remove all packages when ensuring absent

ehelms:
Addresses feedback from previous commit 734b44b8dd2061743f0b12695482ac691682deb5
to not remove ancillary packages and ensure relationship between user
and group.

Attempts to address feedback from https://github.com/theforeman/puppet-qpid/pull/154. This does not currently address https://github.com/theforeman/puppet-qpid/pull/154#discussion_r607744116 as there is discussion around it.

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -47,8 +47,8 @@ class qpid::service {
       notify  => Service['qpidd'],
     }
 
-    if $qpid::ssl {
-      ensure_packages(['iproute'], {ensure => $qpid::ensure})
+    if $qpid::ssl and $qpid::ensure == 'present' {
+      ensure_packages(['iproute'])
       Package['iproute'] -> Systemd::Dropin_file['wait-for-port.conf']
     }
   }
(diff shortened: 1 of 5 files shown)


manifests/service.pp
```diff
@@ -48,7 +48,7 @@
     }
 
     if $qpid::ssl {
```
ekohl, line 50: This should include a check of `$qpid::ensure == 'present'`

spec/classes/qpid_spec.rb
```diff
@@ -67,7 +67,6 @@
         it { is_expected.to contain_class('qpid::service') }
         it { is_expected.to contain_systemd__dropin_file('wait-for-port.conf').with_ensure('absent') }
         it { is_expected.to contain_systemd__service_limits('qpidd.service').with_ensure('absent') }
-        it { is_expected.to contain_package('iproute').with_ensure('absent') }
```
ekohl, line 70: This should still be checked to be in the catalog, just not `with_ensure`.

ekohl (changes requested):

ekohl (commented):
Code wise it looks good. I do wonder about `#32037-2`. What's the -2 part?

ehelms:
> `#32037-2`. What's the -2 part?

A branch naming scheme that made its way into my commit message accidentally.

ekohl (approved):

PR #161, opened 2021-04-13
Refs #32037: Manage /var/lib/qpidd

ehelms:
When the qpidd service starts it creates a directory /var/lib/qpidd/.qpidd
If the qpid is then ensured to be absent, removing the qpidd user
and then re-enabled this directory will have the wrong user id set on it
and get permission denied when trying to start the service.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -33,4 +33,12 @@ class qpid::config
     mode    => '0640',
     content => $qpid::acl_content,
   }
+
+  file { $qpid::data_dir:
+    ensure => bool2str($qpid::ensure == 'present', 'directory', $qpid::ensure),
+    owner  => $qpid::user,
+    group  => $qpid::group,
+    mode   => '0755',
+    force  => true,
+  }
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -67,6 +67,8 @@
 #
 # $ensure::                   Specify to explicitly enable Qpid installs or absent to remove all packages and configs
 #
+# $data_dir::     Location on disk that qpid broker data is stored
+#
 class qpid (
   String $version = $qpid::params::version,
   Boolean $auth = $qpid::params::auth,
@@ -96,6 +98,7 @@ class qpid (
   Boolean $service_ensure = true,
   Optional[Boolean] $service_enable = undef,
   Enum['present', 'absent'] $ensure = 'present',
+  Stdlib::AbsolutePath $data_dir = '/var/lib/qpidd',
 ) inherits qpid::params {
   if $ssl {
     assert_type(Boolean, $ssl_require_client_auth)
--- a/spec/acceptance/qpid_spec.rb
+++ b/spec/acceptance/qpid_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper_acceptance'
 
-describe 'qpid' do
+describe 'qpid', :order => :defined do
   context 'with default parameters' do
     let(:pp) do
       <<-PUPPET
@@ -21,6 +21,10 @@ describe 'qpid' do
 
     describe package('qpid-tools') do
       it { is_expected.to be_installed }
+    end
+
+    describe file('/var/lib/qpidd') do
+      it { is_expected.to be_directory }
     end
   end
 
@@ -46,6 +50,10 @@ describe 'qpid' do
 
     describe package('qpid-tools') do
       it { is_expected.not_to be_installed }
+    end
+
+    describe file('/var/lib/qpidd') do
+      it { is_expected.not_to exist }
     end
   end
 end
--- a/templates/qpidd.conf.erb
+++ b/templates/qpidd.conf.erb
@@ -30,6 +30,7 @@
 <% unless [nil, :undefined, :undef, ''].include?(scope['qpid::acl_content']) -%>
 acl-file=<%= scope['qpid::acl_file'] %>
 <% end %>
+data-dir=<%= scope['qpid::data_dir'] %>
 log-enable=<%= scope['qpid::log_level'] %>
 log-to-syslog=<%= scope['qpid::log_to_syslog'] ? 'yes' : 'no' %>
 auth=<%= scope['qpid::auth'] ? 'yes' : 'no' %>
--- a/spec/classes/qpid_spec.rb
+++ b/spec/classes/qpid_spec.rb
@@ -18,6 +18,7 @@ describe 'qpid' do
         it { is_expected.to contain_class('qpid::config') }
         it 'should create configuration file' do
           verify_exact_contents(catalogue, '/etc/qpid/qpidd.conf', [
+            'data-dir=/var/lib/qpidd',
             'log-enable=error+',
             'log-to-syslog=yes',
             'auth=no'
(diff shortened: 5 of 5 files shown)


spec/acceptance/qpid_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper_acceptance'
 
-describe 'qpid' do
+describe 'qpid', :order => :defined do
```
wbclark, line 3: This is unfamiliar to me.

I would guess it ensures that the tests below run in the order they are defined, so that the `ensure => 'absent'` test is guaranteed to test removal of something that was already installed.

Is that correct and what would be the impact if this change were not included? Does this mean the other tests within that context (e.g. ensureing qpid-tools package is not installed) were not running in the correct order previously?

Are there other `ensure => 'absent'` like tests in our puppet module ecosystem, particularly related to your work about enabling and disabling katello-agent infrastructure, that should receive a similar change?

Thanks @ehelms ... any clarity you can provide around it is helpful.

manifests/config.pp
```diff
@@ -33,4 +33,12 @@
     mode    => '0640',
     content => $qpid::acl_content,
   }
+
+  file { '/var/lib/qpidd':
+    ensure => $qpid::ensure,
+    owner  => $qpid::user,
+    group  => $qpid::group,
+    mode   => '0755',
+    force  => true,
```
wbclark, line 42: :+1: it looks like `force` is strictly necessary with `ensure => 'absent'`.

wbclark (approved):
Tests look good and should confirm the feature is working as intended. This looks to be working and adding value.

I would appreciate a brief explanation about `:order => :defined` in the acceptance test (see inline comment about it) but otherwise it looks good to merge.

wbclark:
I labeled it as a 'bug' for now, as that is consistent with how the need for this change was discovered.

Arguably it should be labeled as a 'feature' instead, because it adds management of an additional resource that an user wouldn't find via any changelog entry at the 'feature' level. Thinking about it, I think 'feature' makes more sense, so I'll update the label again.

spec/acceptance/qpid_spec.rb
ehelms: Exactly. The intent here was to try to enforce setting up qpid and then perform the `ensure => absent` with an installed qpid to ensure that everything worked as expected. The other way to do this would be to try and apply qpid within the absent test block independently. That just felt like overkill when I could use a defined order.

ehelms:
> Arguably it should be labeled as a 'enhancement' instead, because it adds management of an additional resource that an user wouldn't find via any changelog entry at the 'enhancement' level. Thinking about it, I think 'enhancement' makes more sense, so I'll update the label again.

Yea, thats a tough call since without this you cannot apply, ensure absent and then ensure present and expect things to work.

ehelms:
Let's see what @ekohl thinks in case there is a better way to handle (or think about?) this testing to ensure functionality and help prevent regressions.

spec/acceptance/qpid_spec.rb
```diff
@@ -47,5 +47,9 @@ class { 'qpid':
     describe package('qpid-tools') do
       it { is_expected.not_to be_installed }
     end
+
+    describe file('/var/lib/qpidd') do
+      it { is_expected.not_to exit }
```
ekohl, line 52:
```suggestion
      it { is_expected.not_to exist }
```

manifests/config.pp
```diff
@@ -33,4 +33,12 @@
     mode    => '0640',
     content => $qpid::acl_content,
   }
+
+  file { '/var/lib/qpidd':
```
ekohl, line 37: I've tried to avoid hardcoding any paths and still think that's a good pattern to keep. Experience tells me that paths can change, either in SCL or in a new major OS version. That's why I'd prefer to keep it here too.

ekohl (changes requested):
Some implementation comments, but :+1: for the general direction.

manifests/init.pp
```diff
@@ -96,6 +98,7 @@
   Boolean $service_ensure = true,
   Optional[Boolean] $service_enable = undef,
   Enum['present', 'absent'] $ensure = 'present',
+  Stdlib::AbsolutePath $data_dir = '/var/lib/qpidd/.qpidd',
```
ehelms, line 101: The default data directory in qpid-cpp is this path here. This implementation would manage this directory but would not manage the top level directory `/var/lib/qpidd` (which is created by the qpid-cpp RPM).

Options:

 1) Stick with this pattern as that is how things work today
 2) Switch to specifying `/var/lib/qpidd` as the data directory so we can manage and remove it entirely as a single entity. This should not break upgrades given the configuration  would get updated and restart the qpidd service. Any old `.qpidd` data would then get cleaned up anytime that someone set qpid to ensure absent,

templates/qpidd.conf.erb
```diff
@@ -30,6 +30,7 @@
 <% unless [nil, :undefined, :undef, ''].include?(scope['qpid::acl_content']) -%>
 acl-file=<%= scope['qpid::acl_file'] %>
 <% end %>
+data-dir=<%= scope['qpid::data_dir'] %>
```
ehelms, line 33: If we are going to manage the data directory, we should specify it in the configuration for consistency across installs.

manifests/init.pp
ekohl: I'm in favor of using `/var/lib/qpidd`, mostly because that's the home:
```console
# getent passwd qpidd
qpidd:x:998:998:Owner of Qpidd Daemons:/var/lib/qpidd:/sbin/nologin
# rpm -qv qpid-cpp-server-linearstore       
qpid-cpp-server-linearstore-1.39.0-1.el7.x86_64
```

So I'd name it `user_home` and remove that.

ehelms: I made the change to `/var/lib/qpidd` but stuck with `data_dir` as the parameter since that is the Qpid configuration that is being specified.

manifests/config.pp
```diff
@@ -33,4 +33,18 @@
     mode    => '0640',
     content => $qpid::acl_content,
   }
+
+  if $qpid::ensure == 'present' {
+    $_data_dir_ensure = 'directory'
+  } else {
+    $_data_dir_ensure = $qpid::ensure
+  }
```
ekohl, line 41: Sometimes I miss ternaries. This isn't that much better
```suggestion
  $_data_dir_ensure = bool2str($qpid::ensure == 'present', 'directory', $qpid::ensure)
```

What do you think?

ekohl (approved):

ekohl (approved):

ekohl:
@wbclark you approved it in the past, but it changed since. Still think it's ok?

Regarding bug vs enhancement: I agree it's difficult. It's really a bugfix for an enhancement but given the enhancement is unreleased I think I agree it's now an enhancement.

spec/acceptance/qpid_spec.rb
```diff
@@ -47,5 +47,9 @@ class { 'qpid':
     describe package('qpid-tools') do
       it { is_expected.not_to be_installed }
     end
+
+    describe file('/var/lib/qpidd') do
+      it { is_expected.not_to exist }
```
wbclark, line 52: NIT: as long as we are testing that the directory removal works, we might as well also test that we properly added it in the first place. Under the context 'with default parameters' above, you could add a similar test:

```
describe file('/var/lib/qpidd') do
  it { is_expected.to be_directory }
end
```

templates/qpidd.conf.erb
wbclark: :+1: this will be necessary if the user specifies some non-default directory

spec/acceptance/qpid_spec.rb
```diff
@@ -22,6 +22,10 @@
     describe package('qpid-tools') do
       it { is_expected.to be_installed }
     end
+
+    describe file('/var/lib/qpidd') do
+      it { is_expected.to exist }
```
wbclark, line 27: Is it preferable to use `is_expected.to be_directory` ?

wbclark (approved):
Overall this still looks good to me

PR #167, opened 2021-05-23
Add a Rubocop config + update to vp-test 2

ekohl:


spec/classes/qpid_client_spec.rb
```diff
@@ -11,35 +11,35 @@
...
-            'ssl-cert-db=/etc/pki/katello/nssdb',
-            'ssl-cert-password-file=/etc/pki/katello/nssdb/nss_db_password-file',
-            'ssl-cert-name=broker'
-          ])
+                                  'log-enable=error+',
+                                  'ssl-port=5671',
+                                  'ssl-cert-db=/etc/pki/katello/nssdb',
+                                  'ssl-cert-password-file=/etc/pki/katello/nssdb/nss_db_password-file',
+                                  'ssl-cert-name=broker'
+                                ],)
```
ehelms, line 42: I know these are Rubocop rules, but I always disliked how these end up formatting with massive indents when it comes to readability. Just a comment, no action needed.

ekohl: Perhaps this is a rule we want to change. The trailing comma is also a bit weird.

PR #172, opened 2021-08-03
Fix beaker task

ekohl:
In puppetlabs_spec_helper 4.x the beaker task was removed. This loads it from voxpupuli-aceptance.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,10 +1,26 @@
 # This file is managed centrally by modulesync
 #   https://github.com/theforeman/foreman-installer-modulesync
 
-require 'voxpupuli/test/rake'
+# Attempt to load voxupuli-test (which pulls in puppetlabs_spec_helper),
+# otherwise attempt to load it directly.
+begin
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -1,10 +1,26 @@
...
-# We use fixtures in our modules, which is not the default
-task :beaker => 'spec_prep'
+# load optional tasks for acceptance
+# only available if gem group releases is installed
+begin
+  require 'voxpupuli/acceptance/rake'
+rescue LoadError
+else
+  # We use fixtures in our modules, which is not the default
+  task :beaker => 'spec_prep'
```
ehelms, line 22: This is still being called by the looks of the test output

ekohl: Yes, I noticed this too after I sent a bulk of PRs. https://github.com/theforeman/puppet-candlepin/pull/204 is updated and appears to be passing.

ehelms (approved):

PR #174, opened 2021-10-21
Reflect stdlib's ensure_packages default change

ekohl:
ensure_packages() now defaults to ensure => installed where it used to default to present. This reflects it in the tests' expectation.

https://github.com/puppetlabs/puppetlabs-stdlib/commit/77a9c07c9e64f03b12a4d8068941609f44829c8b

### DIFF
--- a/spec/classes/qpid_spec.rb
+++ b/spec/classes/qpid_spec.rb
@@ -182,7 +182,7 @@ describe 'qpid' do
             .that_notifies('Service[qpidd]')
             .that_requires('Package[iproute]')
           is_expected.to contain_package('iproute')
-            .with_ensure('present')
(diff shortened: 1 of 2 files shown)


ekohl:
I just noticed CI didn't run because it was disabled after no changed in the module for a while. That gave a false impression it passed. In the next update cycle I'd like to update our Github workflows. In particular include this bit: https://github.com/voxpupuli/modulesync_config/blob/2349ae417404e6736ec78cb0bb44637e453006a3/moduleroot/.github/workflows/ci.yml.erb#L108-L117

Then we can mark that as a required check. This also gives us the option to enable auto merging once CI completes (and is green).

PR #178, opened 2022-06-08
Drop EL7 support

ehelms:


ekohl:
Somehow the systemd fact gets set to false, but in c34a75fd9d356b1fa2b330346d8eb1e1c2082239 we should have set that to true. It looks like CI didn't ran then which caused us to miss the failure. I'll set up branch protection so it doesn't happen again and look into the failure.

ekohl:
I think it may be down to https://github.com/voxpupuli/voxpupuli-test/pull/80 but locally I have some issues to replicate this. Somehow my custom facts are getting added as strings instead of symbols. I know where it happens now, but I'm thinking about a proper fix without breaking things.

ekohl:
https://github.com/voxpupuli/voxpupuli-test/pull/81 took a while to understand and verify. I'll try to get a release out.

ekohl (approved):
I released voxpupuli-test 5.4.1 and restarted the tests.

